PR #300, opened 2013-08-05
Change httpclient adapter to use thread local variables for thread safety

tbeauvais:
This fixes a threading issue when doing load testing using webmock for our HttpClient service calls using JRuby. The httpclient_adapter was using shared instance variables which caused response corruption.

Oh I noticed before I even started that there were several spec failures. I looked it to it quickly and noticed that calls http://www.example.com/ now results in 200, with actual content. Is someone looking at that?


bblimke:
@tbeauvais Just FYI I have re-introduced your excellent solution to the thread safety problem issue of the httpclient adapter. Thank you! Ruby 1.8.7 is not an issue anymore ;)

PR #384, opened 2014-04-07
Fix query mapper

insoul:
`WebMock::Util::QueryMapper` and `Rack::Utils` are different with nested parameters

``` ruby
WebMock::Util::QueryMapper.values_to_query('one' => ['1','2'])
#=> "one[0]=1&one[1]=2"
WebMock::Util::QueryMapper.query_to_values('one[0]=1&one[1]=2')
#=> {"one"=>["1", "2"]}
WebMock::Util::QueryMapper.query_to_values('one[]=1&one[]=2')
#=> {"one"=>["1", "2"]} # same with 'one[0]=1&one[1]=2'
```

``` ruby
Rack::Utils.build_nested_query('one' => ['1','2'])
#=> "one[]=1&one[]=2"
Rack::Utils.parse_nested_query('one[]=1&one[]=2')
#=> {"one"=>["1", "2"]}
Rack::Utils.parse_nested_query('one[0]=1&one[1]=2')
#=> {"one"=>{"0"=>"1", "1"=>"2"}}
```

I wonder what is correct mapping.

Anyway, when calling  `Curl.get('http://example.com/test?one[]=1&one[]=2')`, WebMock shows tip.

```
You can stub this request with the following snippet:

stub_request(:get, "http://example.com/test?one%5B0%5D=1&one%5B1%5D=2").
  to_return(:status => 200, :body => "", :headers => {})
```

After decoding url of tip above, query is converted to `one[0]=1&one[1]=2`
This is confusing. So I fixed only this case.

``` ruby
WebMock::Util::QueryMapper.values_to_query('one' => ['1','2'])
#=> "one%5B%5D=1&one%5B%5D=2"
```

### DIFF
--- a/lib/webmock/util/query_mapper.rb
+++ b/lib/webmock/util/query_mapper.rb
@@ -133,8 +133,10 @@ module WebMock::Util
         new_query_values = new_query_values.inject([]) do |object, (key, value)|
           key = key.to_s if key.kind_of?(Symbol) || key.nil?
 
-          if value.respond_to?(:each)
+          if value.is_a?(Array)
             value.each { |v| object << [key.to_s + '[]', v] }
+          elsif value.is_a?(Hash)
+            value.each { |k, v| object << ["#{key.to_s}[#{k}]", v]}
           else
             object << [key.to_s, value]
           end
--- a/spec/unit/util/query_mapper_spec.rb
+++ b/spec/unit/util/query_mapper_spec.rb
@@ -33,4 +33,11 @@ describe WebMock::Util::QueryMapper do
     expect(query_mapper.values_to_query values).to eq query
     expect(query_mapper.query_to_values query).to eq values
   end
+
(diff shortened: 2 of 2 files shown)


mikegee:
I'm running into this error too: `ArgumentError: comparison of Array with Array failed`.

Here's a sample of the code from my test suite (I tried to reduce it to the minimal repro):

```ruby
params = { tags: [ { foo: :bar, baz: :quux } ] }
url = "http://example.com?#{params.to_param}"
stub_request :get, url
```

I put this test case in webmock's suite and it fails:

```ruby
it 'does not raise',focus:true do
  url = 'http://example.com?tags%5B%5D%5Bbaz%5D=quux&tags%5B%5D%5Bfoo%5D=bar'
  expect { stub_request(:get, url) }.to_not raise_error
end
```

I was about to make a PR, but webmock's test suite is kinda daunting.

PR #398, opened 2014-06-19
Add ability to allow net connect inside a block

lserman:
If a block is given to allow_net_connect!, allow net connection within
it and then disable_net_connect! automatically.

I think this syntax has better aesthetics than explicitly calling
disable_net_connect!, as it nicely groups allowed code within the
block.

### DIFF
--- a/lib/webmock/config.rb
+++ b/lib/webmock/config.rb
@@ -6,5 +6,15 @@ module WebMock
     attr_accessor :allow_localhost
     attr_accessor :allow
     attr_accessor :net_http_connect_on_start
+
+    def to_h
+      {
+        allow_net_connect: allow_net_connect,
+        allow_localhost: allow_localhost,
+        allow: allow,
+        net_http_connect_on_start: net_http_connect_on_start
+      }
+    end
+
   end
 end
--- a/lib/webmock/webmock.rb
+++ b/lib/webmock/webmock.rb
@@ -42,11 +42,12 @@ module WebMock
   end
 
   def self.allow_net_connect!(options = {})
+    state_snapshot = Config.instance.to_h
     Config.instance.allow_net_connect = true
     Config.instance.net_http_connect_on_start = options[:net_http_connect_on_start]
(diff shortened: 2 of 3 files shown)


hovissimo:
I'm not in a position to provide a PR for this, but I'll add my name to the hat for wanting this feature.

My use case is that I'm a new engineer on my team, and the state of WebMock isn't guaranteed as I enter my tests. However, further tests do depend on the state of WebMock, so I want to make sure I restore that state as my tests exit.  In a perfect world we would all have time to make beautiful, consistent tests and I wouldn't need this, but as it is I wish I had a feature like this.

bblimke:
@hovissimo in your case it's not only about allow_net_connect but about the state of the whole WebMock. WebMock state also includes stubs or callbacks.
Making allow_net_connect accept a block would only be a partial solution.

hovissimo:
Okay.  I'm fairly new to Ruby/Rails, maybe you can help me reformulate my request?

It seems like an extremely handy thing to be able to configure Webmock for a block.

Something like:
```
Webmock.with_options({ foo: 'bar', qux: @quux }) do
  # Test all the things
end
```

This structure would not only be hand for manually configuring Webmock for individual tests, but I think this would also become a very sane and trustworthy way to configure categories of tests (via ancestor classes or what have you).  I'm afraid I don't know a ton of under-the-hood Ruby, so I may be a little confused.

bblimke:
@hovissimo in WebMock stubs, allowed network connection patterns, requests are all globals.

They get reset before each test/spec by default.

In order to support block, we would have have to keep contexts.

When existing block WebMock should return to previous context,
which means it would have to remember all it's state.

What happens when we have some stubs declared and we enter a block? Do existing settings propagate or not?

When happens when we declare stubs inside the block and then we exist block? Do stubs still exist?

hovissimo:
It sounds like Webmock's architecture is far enough from my idea that it's not worth pursuing.

PR #416, opened 2014-10-06
Nested Hash Partial Matching

BrandonJoyce356:
I would really like to have partial matching on nested hashes.  I've written a failing test here to show what I mean.  Would you like me to implement this?


presidentJFK:
closing on request. Thanks for submitting.

PR #418, opened 2014-10-17
Ensure httpclient 2.5.0 and greater is being used

zzak:


zzak:
@bblimke Using HTTPClient < 2.5 in development could affect https connections for anyone inheriting the defaults of OpenSSL for these connections.

bblimke:
"Using HTTPClient < 2.5 in development could affect https connections for anyone inheriting the defaults of OpenSSL for these connections." - is this still true if people don't use WebMock?

PR #436, opened 2014-11-10
Ignore ArgumentError during sort of non-comparable values

magneland:
Since the sort is optional anyways, it’s better to just continue.


presidentJFK:
@magneland so I took a look at this and like @bblimke said Ruby 1.8.7 NoMethodError and later rubies, raises a ArgumentError when trying to sort arrays of hashes.

So It seems that we just want to rescue both.

So I made a PR #514 that does that and CI now passes. So I am going to close this Pull Request and merge the other in. Thanks for taking time to submit this!

magneland:
Thank you, Mr. President!

PR #441, opened 2014-12-01
Fix Json parse problem with multibyte characters

VincentZhao:
When dealing with multibyte characters, sometimes inappropriate white spaces are inserted.
This PR fixes this.

### DIFF
--- a/lib/webmock/util/json.rb
+++ b/lib/webmock/util/json.rb
@@ -42,10 +42,16 @@ module WebMock
           json.gsub(/\\\//, '/')
         else
           left_pos  = [-1].push(*marks)
-          right_pos = marks << json.length
+          right_pos = marks << json.bytesize
           output    = []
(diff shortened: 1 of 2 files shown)


presidentJFK:
So I have recreated this error locally and am currently exploring what is exactly happening.

It seems to be an issue inside the Syck Library, but not sure exactly what yet.

presidentJFK:
So @VincentZhao I have finally figured this out to be a problem with ruby 1.9.2.

It has to do with 1.9.2. not having String#byteslice, 
and when accessing characters with the string[1..2] syntax
1.8.7 returns the integer value
1.9.2 returns the string.

So I have fixed this by converting the json string to binary encoding and then back to UTF-8. So I am going to close this branch and merge in #512.

Thanks for submitting this PR and adding this functionality!

PR #447, opened 2015-01-26
Allow RSpec like syntax

frankywahl:
This PR allows a more intuitive approach when using RSpec.
Since double and mock in RSpec already permit a `.and_return` syntax,
mimic this behaviour in webmock for easier reading and integration.

### DIFF
--- a/lib/webmock/request_stub.rb
+++ b/lib/webmock/request_stub.rb
@@ -22,6 +22,7 @@ module WebMock
       end
       self
     end
+    alias_method :and_return, :to_return
 
     def to_rack(app, options={})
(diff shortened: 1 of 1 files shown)


bblimke:
Nice addition. Unfortunately it broke Travis build on all platforms.

frankywahl:
Yeah, saw that. Was a typo...
Any reason `jruby-18mode` is an allowed failure but not `jruby-19mode`?
(I've never actually worked with jruby, and don't know if it may be a reason it's not passing...)

bblimke:
I think it's randomly failing test.

jruby-18mode was failing so often I had to put it into allowed failures section. jruby 1.9 usually works.

bblimke:
All good. Cheers!

PR #448, opened 2015-01-26
Remove unnecessary @query_params.nil? check

fishermand46:
In this branch, we already know that @query_params is not nil. Looks like this was the result of a copy-pasta. :P


bblimke:
It's obsolete indeed :)

PR #456, opened 2015-02-19
Detect presence of HTTP gem a little more cleanly.

simonrussell:
Currently the loaders assume that the http gem is there as long as "require 'http'" doesn't fail -- which doesn't work well for codebases with a file called "http".

This pull request resolves that by making the HTTP gem detection work more like how all the other detectors work.

### DIFF
--- a/lib/webmock/http_lib_adapters/http_gem_adapter.rb
+++ b/lib/webmock/http_lib_adapters/http_gem_adapter.rb
@@ -1,11 +1,10 @@
 begin
   require "http"
-  __http_gem_found__ = true
 rescue LoadError
-  __http_gem_found__ = false
+  # HTTP gem not found
 end
 
-if __http_gem_found__
+if defined?(HTTP) && defined?(HTTP::VERSION)
   WebMock::VersionChecker.new("HTTP Gem", HTTP::VERSION, "0.6.0").check_version!
 
   module WebMock


bblimke:
It's very likely that if there is a file called http.rb then the class defined inside is called HTTP, but less likely that there is HTTP::VERSION . It looks good. Thanks!

simonrussell:
Thanks for merging, this was a real issue for us; our module is called "Http" (Guardian-style acronyms.)  But yeah, I guess no real guarantee that 'require "http"' defines a class/module at all.

Good project, by the way; thanks for your work.

lib/webmock/http_lib_adapters/http_gem_adapter.rb
```diff
@@ -1,11 +1,10 @@
 begin
   require "http"
-  __http_gem_found__ = true
 rescue LoadError
-  __http_gem_found__ = false
+  # HTTP gem not found
 end
 
-if __http_gem_found__
+if defined?(HTTP) && defined?(HTTP::VERSION)
```
beedub: Can't this be truncated to `if defined?(HTTP::VERSION)`?

simonrussell: You're right actually, I assumed it only checked the last thing in the chain (HTTP isn't defined in my project) but it seems that it's fine with any of the things being undefined.

PR #472, opened 2015-05-01
fix spelling mistake stib -> stub

presidentJFK:


bblimke:
@presidentJFK thank you! :)

PR #473, opened 2015-05-01
Fix travis

presidentJFK:


presidentJFK:
This update in the http gem broke 1.8.7 and ree support
https://github.com/httprb/http.rb/commit/110a3380e715446f44ca2695b8e18a57e2abc3bd

presidentJFK:
1.9.3 is failing calling a protected params method, which I believe was introduced with this commit.
https://github.com/httprb/http.rb/blob/91a2092bf51f66da185a8b3cdb7351acb980fafd/lib/http/options.rb#L68

bblimke:
@presidentJFK thank you very much for fixing the build!

PR #478, opened 2015-05-10
Fix Json parse problem with multibyte characters

presidentJFK:
this is rebased on master, to fix the CI build


presidentJFK:
I was hoping for an easy win, and for the other CI fixes to fix this, but alas no.

So I will have to dig into this encoding bussiness later.

PR #480, opened 2015-05-12
show a hash diff for requests that have stubs with body set;

yurivm:
I am an avid user of Webmock and I use it in specs to stub requests specifying body as a JSON string, like this:
```ruby
stub_request(:post, url).
         with(:body => '{"request":"something","reference":"REF123","store_data":"true","email":"testme@test.com","test":"test2","test3":"test4"}',
              :headers => headers)
```
If I don't get it right and send a request with a different body, Webmock produces an error messages along the lines of:

WebMock::NetConnectNotAllowedError:
Real HTTP connections are disabled. Unregistered request: POST http://example.com with body '{"request":"something","reference":"REF-123","store_data":"true","email":"testme@test.com","test2":"test","test4":"test5"}'
registered request stubs:

```ruby
stub_request(:post, url).
         with(:body => '{"request":"something","reference":"REF123","store_data":"true","email":"testme@test.com","test":"test2","test3":"test4", }',
````

And then I have to manually compare both strings to find out that I haven't provided the "test" key and used "test2" instead. 
In my case bodies are much longer and comparing them takes a while and is very error-prone.
With this PR, webmock will include the body diff obtained via [hashdiff](https://github.com/liufengyun/hashdiff)  for each registered stub like this:

Body diff:
        [["-", "test", "test2"], ["+", "test4", "test5"]] 

So you clearly see that your request didn't have the "test" key with "test2" value but did have "test4" with "test5" on top.
This makes it much simpler to find what's missing or wrong.

### DIFF
--- a/lib/webmock.rb
+++ b/lib/webmock.rb
@@ -28,6 +28,8 @@ require 'webmock/response'
 require 'webmock/rack_response'
 
 require 'webmock/stub_request_snippet'
+require 'webmock/request_signature_snippet'
+require 'webmock/request_body_diff'
 
 require 'webmock/assertion_failure'
 require 'webmock/request_execution_verifier'
--- a/lib/webmock/config.rb
+++ b/lib/webmock/config.rb
@@ -4,6 +4,7 @@ module WebMock
 
     def initialize
       @show_stubbing_instructions = true
+      @show_body_diff = true
     end
 
     attr_accessor :allow_net_connect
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

**Comment by presidentJFK** [comment]
Thanks for the pull request! I am going try and review it later tonight. I do however notice that CI is failing because it can't load JSON library.

```
request_body_diff.rb:2:in `require': cannot load such file -- JSON (LoadError)
```

I opened a branch where I will experiment with getting the CI passing. But if you could take a look that be great.

presidentJFK:
I changed JSON to json and that fixed 4 failing builds. Heres the PR https://github.com/bblimke/webmock/pull/481

But there are still a lot of failures, that I will look into later.

presidentJFK:
closing this now that #481 is merged in

PR #481, opened 2015-05-12
Net connection show body diff

presidentJFK:


yurivm:
Thanks, I'll take a look now

yurivm:
Hmm, I've tried both 2.1.0 and 2.2.1 and was unable to reproduce failed specs locally. I did remove an unnecessary require that I had overlooked in #480 .
How can I help you fix the builds? Does travis take any special steps I might have missed?

presidentJFK:
There shouldn't be any special setup, but I am very new to contributing to this project, and have just started getting more familiar with the CI setup. 

I am going to start digging into the log https://s3.amazonaws.com/archive.travis-ci.org/jobs/62266252/log.txt

And maybe if I recreate using the exact same dependencies, I can get it failing local.

But your guess is as good as mine, as to what the problem is. If I mind anything I'll post it here.

yurivm:
I've created a virtual machine with Ubuntu 12.04 (which is what Travis seems to use) and followed all the steps in the log.
I had to install libcurl4-openssl-dev to get curb to compile its curl extension, but other than that, no issues and no rspec errors either. Same gem versions and the same ruby version as in the log.

presidentJFK:
Thanks for trying that out, but bummer it didn't help!

I know there are warnings about the addressable dependency, and I see 2 of the builds are failing, because of something involving addressable. So I am going to start exploring that.

There is also a chance there might be intermittent failures on Travis, which I haven't looked into yet.

yurivm:
I was able to reproduce failing specs with Ruby 2.0 - I'll have a closer look this weekend

presidentJFK:
@yurivm The specs that were failing, did not have anything to do with this feature and are now fixed. So I have rebased this branch of these changes on master, and everything is green on CI now.

I'm now just reviewing everything to make sure I understand all the changes (more to help my own knowledge as I am still becoming familiar with all the areas of WebMock) and we can get this merged in!

presidentJFK:
@yurivm I took a look at everything and it all looks good. 

I wrote some unit tests for RequestBodyDiff and RequestSignatureSnippet, to just have better test coverage and make myself more familiar with the code.

And as a result of this test writing, I ended up refactoring RequestBodyDiff's internals to read a little simpler, and I also restricted RequestSignatureSnippet's API to just #stubbing_instructions and #request_stubs.

I also ma

presidentJFK:
@yurivm I'm going to ahead and merge this in. Thanks for the feature!

PR #487, opened 2015-06-12
Failing spec to expose problem mentioned in #384

mikegee:


mikegee:
I suspect this spec doesn't belong here. And, I haven't looked into the implementation yet. Helpful suggestions are welcome.

mikegee:
`lib/webmock/util/query_mapper.rb:199` is:

    new_query_values.sort!

In my example, `new_query_values` is:

    [["tags[]", {"baz"=>"quux"}], ["tags[]", {"foo"=>"bar"}]]

Sorting this is problematic because:

    {'baz'=>'quux'} <=> {'foo'=>'bar'} # => nil

`Array#sort!` doesn't like `nil`s.

mikegee:
I rewrote the spec and moved it to a better home.

bblimke:
Thank you @mikegee This should help when developing the fix.

PR #492, opened 2015-07-18
Fix README example with bad copy & paste

esambo:
hash keys should be unique


bblimke:
@esambo thank you. The previous code was indeed confusing.

PR #493, opened 2015-07-18
Fix exception raised when keys are symbols

ramontayag:
When symbols are keys, `QueryMapper#to_query` blows up with `TypeError: can't dup Symbol`.


bblimke:
@ramontayag thanks!

PR #495, opened 2015-07-21
Fix most tests on ruby 2.x and rspec raise_error warnings

tmertens:
This fixes the majority of failures in ruby 2.x in what should be a backwards-compatible way.  It also fixes rspec warnings about unbounded use of the `raise_error` matcher.

There are still two failures I see locally with ruby 2.1.5 which appear legitimate and may reflect changed behavior in newer versions of library dependencies such as httpclient.  The outstanding failures include:

```
spec/acceptance/httpclient/httpclient_spec.rb:126
spec/acceptance/http_rb/http_rb_spec.rb:57
spec/acceptance/http_rb/http_rb_spec_helper.rb:5
```

The failure coming from http_rb_spec_helper is related to https and may be due to missing certs when running the specs, so I haven't spent time looking into it.  The first two look like potentially legitimate failures.

### DIFF
--- a/spec/acceptance/http_rb/http_rb_spec.rb
+++ b/spec/acceptance/http_rb/http_rb_spec.rb
@@ -19,7 +19,7 @@ describe "HTTP.rb" do
 
     it "fails if body was already streamed" do
       response.body.to_s
-      expect { response.body.readpartial 1 }.to raise_error
+      expect { response.body.readpartial 1 }.to raise_error ArgumentError
     end
   end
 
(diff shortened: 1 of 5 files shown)


bblimke:
@tmertens what was the reason to change `eql('200')` to `match /200/`? To make it match numbers as well as strings? 
Webmock should have unified way of returning status codes and these are strings. Therefore if one of the adapters started returning a number (due to changed in underlying http client library), then the change should happen in the http lib adapter to return correct string status code.

tmertens:
@bblimke The change is because newer versions of ruby - I suspect, more specifically, newer versions of some gem dependency - has changed the response status, to include the response code & message, e.g. '200' => '200 OK'.  So, the updated matchers will match the message for both new and old versions of whatever library has changed.  I didn't dig deep into it, but you can see on recent builds of other branches this failure has appeared on newer versions of ruby.

bblimke:
@tmertens thanks for clarifying. The tests (these related to `200`) should not be changed. Instead the `httpclient_spec_helper` should be adapted to return '200'  as status.

presidentJFK:
closing after #509 being merged in.

Issue #496, opened 2015-07-21
Should not depend on safe_yaml


spjsschl:
safe_yaml changes how the YAML parser that comes with ruby behaves. By pulling in safe_yaml (via crack) webmock causes inconsistencies between parsing a YAML file in production and test environments. A possible workaround is to require safe_yaml in production. Yet I feel webmock should not dictate how the app parses YAML.

bblimke:
@spjsschl thanks for reporting that.

As I can see safe_yaml is only used by Crack json parser https://github.com/jnunemaker/crack/blob/master/lib/crack/json.rb
WebMock uses own JSON parser nowadays (extracted from previous version of Crack) 

Perhaps it's enough if WebMock requires 'crack/xml', instead of 'crack' and the problem will be solved.

spjsschl:
That sounds, like it would work.

bblimke:
@spjsschl great. I'll do it when I find some time, unless you fancy a pull request :)

spjsschl:
I added a PR: https://github.com/bblimke/webmock/pull/497

FWIW, the same 39 tests are failing on my local machine before and after my change

PR #497, opened 2015-08-05
require only the necessary parts of crack to avoid pulling in safe_yaml

spjsschl:
Fixes https://github.com/bblimke/webmock/issues/496


bblimke:
Thanks!

spjsschl:
thanks for merging. Is it possible to release a patch version that has these changes?

Issue #451, opened 2015-02-03
Boolean values need to be passed as strings to query ("true" instead of true)


riggy:
Hey,

I find it strange since you can actually pass a hash of options as a stubbed query, like
   
    stub_request(:post, "http://something.com").with(query: {some_boolean_option: true})

but the actual url is registered like this:

    http://something.com?some_boolean_option # yeah, nothing here...

It seems that boolean values are not converted to string when constructing stubbed url. Passing value as string works well. It's pretty annoying as it breaks consistency of my tests. Is it intentional or is it just an error or whatever?

bblimke:
It looks like a bug. All query values should be strings, since there is no other way to transfer these via query string.

presidentJFK:
closing now that #513 has been merged in.

PR #498, opened 2015-08-10
fix for #451

riggy:
It only failed for true values, worked fine with false. Removing those two lines did the trick for me.


bblimke:
Does it mean it worked fine for `false`, just not for true? Would be good to have a spec covering that.

presidentJFK:
@bblimke @riggy before this change, false was returning a=false, like it should. This fix does indeed make both true and false behave like that.

I updated a test case for this to check for true and false, and now CI is all green.

So everything looks good, and I am going to close this PR and merge in my updated branch #513.

PR #499, opened 2015-08-17
Turn off sudo

sinisterstuf:
Travis's new container-based infrastructure requires you to explicitly turn off `sudo`. No commands in this build use it, so compliance is as simple as specifying `sudo: false` in .travis.yml


bblimke:
@sinisterstuf thank you!

Issue #454, opened 2015-02-10
No support for Manticore


ebenoist:
Currently Webmock lacks support for the [manticore](https://github.com/cheald/manticore) HTTP client library as it is implemented with the Apache Commons HTTP library (and is thus JRuby only). A recent PR was merged into the manticore library to support [regexps](https://github.com/cheald/manticore/pull/16), which may allow for webmock support. Though other changes may need to be made to allow manticore's stubbing support to have parity with webmock.

I am interested in adding this support if the maintainer would be interested, but wanted your advice on approaches. I imagine it would involve adding a new http library adapter and following something like the typhoeus adapter pattern while swapping out the typhoeus mocks for manticore's stubbing support.

Any advice would be appreciated.

bblimke:
In order to add new manticore client support you need to add `manticore_spec.rb` and `manticore_spec_helper.rb` inside `spec/acceptance/manticore/`based on examples from other libs.
As a base the spec needs to `include_context "with WebMock"`. 

Then you can proceed with implementing the adapter. When all specs are passing, the support is complete.

Since each new library adds maintenance overhead to WebMock, it's best to keep WebMock support only for the most popular clients. Do you think manticore is popular enough?

jkingdon:
As for "only for the most popular clients", manticore seems the leading candidate at [Square](http://square.github.io/) to wean us off `Net::HTTP` for jruby apps. I believe the reasons have to do primarily with the "robust" in "Manticore is a fast, robust . . ." at https://github.com/cheald/manticore (connection pooling and so on) but I could dig up more details if needed.

jkingdon:
Hmm, that particular choice seems to be out of date (in favor of https://github.com/httprb/http.rb ). That's just one data point, I don't really know about manticore's popularity in general.

presidentJFK:
closing now that #523 is merged in

PR #505, opened 2015-08-21
Add manticore adapter

mikeknep:
Adds an adapter for [Manticore](https://github.com/cheald/manticore), addressing #454.

I don't know if there is any functionality unique to Manticore that warrants additional tests beyond the shared "with WebMock" suite. @cheald, any thoughts?

Manticore specs pass locally on JRuby 1.7.8 and JRuby 9.0.0.0. Running the full suite on both JRuby versions yields 36 failing http_rb_adapter specs (this is also the case on master branch).

I noticed that each adapter's spec_helper is handling array values in headers in unique but similar ways. I think a helper could be added to `WebMock::Util::Headers` to address this duplication. However, rather than add that and refactor across libraries now, I figured it'd be better to keep this PR focused exclusively on Manticore. (I bring this up because this PR adds duplicate `join_array_values` methods in the adapter and spec_helper.)

### DIFF
--- a/lib/webmock/http_lib_adapters/manticore_adapter.rb
+++ b/lib/webmock/http_lib_adapters/manticore_adapter.rb
@@ -1,0 +1,104 @@
+begin
+  require 'manticore'
+rescue LoadError
+  # manticore not found
+end
+
+if defined?(Manticore)
+  module WebMock
+    module HttpLibAdapters
+      class ManticoreAdapter < HttpLibAdapter
+        adapter_for :manticore
+
+        OriginalManticoreClient = Manticore::Client
+
+        def self.enable!
+          Manticore.send(:remove_const, :Client)
+          Manticore.send(:const_set, :Client, WebMockManticoreClient)
+        end
+
+        def self.disable!
+          Manticore.send(:remove_const, :Client)
+          Manticore.send(:const_set, :Client, OriginalManticoreClient)
+        end
+
+        class WebMockManticoreClient < Manticore::Client
+          def http(method, uri, options = {})
+            @method = method
+            @uri = uri
+            @options = options
+            super(method, WebMock::Util::URI.normalize_uri(uri).to_s, format_options(options))
+          end
+
+          private
+
+          def format_options(options)
+            return options unless headers = options[:headers]
+
+            options.merge(:headers => join_array_values(headers))
+          end
+
+          def join_array_values(headers)
+            headers.reduce({}) do |h, (k,v)|
+              v = v.join(', ') if v.is_a?(Array)
+              h.merge(k => v)
+            end
+          end
+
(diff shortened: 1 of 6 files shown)


lib/webmock/http_lib_adapters/manticore_adapter.rb
```diff
@@ -0,0 +1,104 @@
...
+          end
+
+          def join_array_values(headers)
+            headers.reduce({}) do |h, (k,v)|
+              v = v.join(', ') if v.is_a?(Array)
+              h.merge(k => v)
+            end
+          end
+
+          def response_object_for(client, request, context, &block)
```
cheald: This is a private method that doesn't seem to be called anywhere; is this correct?

cheald: *sigh* nevermind, it's overriding a Manticore method. I should go make :coffee:

cheald:
To be honest I'm probably the wrong person to evaluate the feature coverage; I haven't used WebMock since the early Rails 2 days :)

Manticore does expose an awful lot of functionality, though, but it tries to have a relatively sane and compact (read: testable) interface. I'm happy to advise on specific examples if the WebMock maintainers can enumerate some potentially questionable cases.

mikeknep:
Update: added support for stubbing requests made via Manticore's facade methods (i.e. calling `Manticore.get(url)` instead of `client = Manticore::Client.new; client.get(url)`).

@bblimke any thoughts on this getting merged in?

bblimke:
@mikeknep I'm won't be able to take care of this until 12th September. @presidentJFK perhaps you fancy looking after it?

presidentJFK:
I don't have any experience using Manticore, so my help might be a little slow. But if no one gets this sooner, I'll take a look once I finish some other Webmock todos.

mikeknep:
Checking in on this. @bblimke or @presidentJFK is there anything you need from me or that I can help with for this PR?

presidentJFK:
@mikeknep this is next up on my list and I will looking at this week, probably starting tonight. 

So I will be hitting you up for questions/feedback soon. Thanks!

bblimke:
Thanks. Everything looks good!

It's only mandatory for each adapter to pass "with WebMock" examples. Additional specs are only needed if given http client provides additional functionalities not covered by "with WebMock" specs.

@mikeknep I hope you can help with Manticore adapter maintenance in case Manticore changes? :)

mikeknep:
Awesome, glad to hear it! And I'll certainly help check out any problems with it in the future.

presidentJFK:
@mikeknep
I open a new Pull Request #523 the fix the CI Test Suite.

presidentJFK:
closing this now that #523 is merged in. Thanks @mikeknep!

Issue #503, opened 2015-08-20
fix failing tests


sinisterstuf:
It is difficult to see whether the project is stable and whether pull requests should be merged because currently the tests fail. Turning off sudo with PR #499 was a simple way to get a lot of them running. I had a look at the rest but unfortunately they aren't as simple and I don't know Ruby well enough. One problem seems to be that the status responses are now not just numbers, i.e. `200 OK` instead of `200`. Issue #494 is possibly related.

presidentJFK:
Thanks for submitting an issue, however I am going to close this as a duplicate of #494.

We are working on this though. I just opened PR #507 built off of @mikeknep PR #506 attempting to fix this issue. We now just need to do some more research into why are all the failures are happening exactly and what's the optimal solution.

So hopefully this won't be an issue anymore soon, but If you have any extra time, any help would be greatly appreciated!

Issue #494, opened 2015-07-21
specs out of date


tmertens:
The readme for webmock states to use the command `bundle exec rspec spec/**/*_spec.rb` (with rvm across ruby versions).  However, this pattern omits a large swath of tests.  Using the aforementioned pattern, only 241 specs run.

If you instead run `bundle exec rake`, 3941 examples are executed, with 40 failures reported (ruby 2.1.5).  Furthermore, this is the command used on travis, at least for newer ver

There are a couple issues here:

1. The readme is either out of date, or the tests are.  The readme should be updated and tests should be updated to either pass or if the tests are no longer valid, they should be deleted.  It appears that the assertions are valid for older versions of ruby, but not newer ones (see below).  It seems in the interest of forward progress to make these tests work on newer / supported versions of ruby, or fix them to pass on all versions.

2. The ruby versions listed in the documentation do not align with the versions on travis, and omits all of the newer versions (2.x).  All of the versions listed on the readme are no longer maintained and some/most do not install out of the box (on OS X 10.10) using rvm without additional work.  Once installed, there are additional challenges in getting webmock gem dependencies installed on older versions of ruby.  For ease of development, I would personally recommend dropping the requirement for running tests locally on old ruby versions and stick to 2.x+ (or 1.9.3 at the earliest), and let travis pick up the slack for older versions if there are plans to actually continue support for long abandoned versions of ruby.

Is there any direction on the above two issues? I may be able to work up a PR to address some of the test/documentation issues, but want to get direction from the maintainers first to avoid wasted effort if this does not align with your roadmap.

bblimke:
@tmertens thank you for opening this issue.

All specs should be executed, therefore `bundle exec rake` is the correct command to run it.

I have been keeping support for Ruby 1.8.7 since it didn't require any additional effort and WebMock was easily able to support these.

I'm keen to keep support for Ruby 1.8.7, unless it makes development a pain.
Could you please provide more info on problems when installing Ruby 1.9.2 and 1.8.7?
I'm using Rbenv and had no issues with it.

presidentJFK:
On the topic of dropping support for 1.8.7. Addressable has dropped support of 1.8.x as of 2.3.8
https://github.com/sporkmonger/addressable/blob/master/CHANGELOG.md

we have an open >= 2.3.6 in the gemspec so 2.3.8 is what is installed for 1.8.7 running on travis (and passing now!)

We might want to lock down the version, which I'm trying out on different ruby versions now.
[...]

PR #506, opened 2015-08-21
[WIP] Fix specs

mikeknep:
## Work in progress - fixing specs

State of the world at d72af3d:
- Specs all pass on MRI 2.2.2 and JRuby 9000
- Two failures on MRI 1.9.3
- Three failures on JRuby 1.7.8

These are the only four versions I've been testing locally. I use rbenv instead of rvm and haven't implemented a way to run all specs across multiple versions in this project with my setup.

### Problematic specs
`spec/acceptance/http_rb/http_rb_spec.rb:57`
- 2-2-2 versions: `response.uri` is an `Addressable::URI` object
- 1.9.3 versions: `response.uri` is a `URI::HTTP` object

The spec could just call `#to_s` on both to pass, but that seems like a cop-out.

`spec/acceptance/httpclient/httpclient_spec.rb:121`
- 2.2.2 versions
  - stubbing cookie header in first only => one request_signature has cookie header, other does not
  - stubbing no headers => green
  - stubbing cookie header in both => green
- 1.9.3 versions
  - stubbing cookie header in first only => green
  - stubbing no headers => one request_signature has no headers, the other has a cookie header
  - stubbing cookie header in both => one request_signature has single cookie header, the other has array of duplicate cookie headers

cc #494, #503

### DIFF
--- a/spec/acceptance/httpclient/httpclient_spec.rb
+++ b/spec/acceptance/httpclient/httpclient_spec.rb
@@ -113,13 +113,8 @@ describe "HTTPClient" do
         nil # to let the request be made for real
       end
 
-      # To make two requests that have the same request signature, the headers must match.
-      # Since the webmock server has a Set-Cookie header, the 2nd request will automatically
-      # include a Cookie header (due to how httpclient works), so we have to set the header
-      # manually on the first request but not on the 2nd request.
-      http_request(:get, webmock_server_url, :client => client,
-                         :headers => { "Cookie" => "bar=; foo=" })
-      http_request(:get, webmock_server_url, :client => client)
+      http_request(:get, webmock_server_url, :client => client, :headers => { "Cookie" => "bar=; foo=" })
+      http_request(:get, webmock_server_url, :client => client, :headers => { "Cookie" => "bar=; foo=" })
 
       expect(request_signatures.size).to eq(2)
       # Verify the request signatures were identical as needed by this example
--- a/spec/acceptance/http_rb/http_rb_spec.rb
+++ b/spec/acceptance/http_rb/http_rb_spec.rb
@@ -49,7 +49,7 @@ describe "HTTP.rb" do
   end
 
   it "restores request uri on replayed response object" do
-    uri = URI "http://example.com/foo"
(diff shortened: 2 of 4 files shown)


presidentJFK:
Thanks for submitting this. I'm taking a look now, and will hopefully figure out how we can incorporate your updates to get back to a green suite!

spec/acceptance/httpclient/httpclient_spec.rb
```diff
@@ -113,13 +113,8 @@ def filter_response(request, response)
...
       end
 
-      # To make two requests that have the same request signature, the headers must match.
-      # Since the webmock server has a Set-Cookie header, the 2nd request will automatically
-      # include a Cookie header (due to how httpclient works), so we have to set the header
-      # manually on the first request but not on the 2nd request.
-      http_request(:get, webmock_server_url, :client => client,
-                         :headers => { "Cookie" => "bar=; foo=" })
-      http_request(:get, webmock_server_url, :client => client)
+      http_request(:get, webmock_server_url, :client => client, :headers => { "Cookie" => "bar=; foo=" })
```
presidentJFK: This looks like it is actually an order dependent test issue. As the spec is passing if run in isolation, or even if you just run this whole file, but fails in the whole suite.

Where as your change passes in the whole suite, but fails in isolation.
I'm trying to track down the issue now, but if please feel free to jump in, you have any ideas.

presidentJFK: looks like I spoke too soon, I have now realized this fails if I run without bundle install, but works with. And the way I have my vim setup to run focused tests uses bundle exec, but I was running the whole suite with just rspec.

So I'm now trying to track down the dependency difference that is causing this to fail.

presidentJFK: So it looks like httpclient made changes around exactly this functionality in 2.6.0
https://github.com/nahi/httpclient/blob/master/CHANGELOG.md

So I am still trying to grok all the changes, reasons, implications etc. And see if we can use 2.6.0 for all ruby versions.

PR #507, opened 2015-08-29
Adjusting dependencies to fix specs

presidentJFK:


presidentJFK:
So @mikeknep I took your pull request and made the smallest changes I could figure out to get a green CI build. But, I am still exploring what is happening in each case, and trying to figure out a pull request we can merge to return a green suite.

Relevant information:

* Changes in 2.6.0 from [httpclient](https://github.com/nahi/httpclient) [CHANGELOG](https://github.com/nahi/httpclient/blob/master/CHANGELOG.md)

* Update 0.8.9 (2015-05-11) from [http](https://github.com/httprb/http) [CHANGELOG](https://github.com/httprb/http/blob/master/CHANGES.md)

* And I am not sure of how [addressable](https://github.com/sporkmonger/addressable) fits into all of this,  or if maybe changes in with the Standard Library URI across Ruby Versions is causing problems.

Any feedback help is greatly appreciated.

presidentJFK:
I am looking into what has caused spec/acceptance/http_rb/http_rb_spec.rb to start to fail.

And it looks like this commit is the culprit https://github.com/httprb/http/commit/3a06852cad71d8611d2a17fdfca8da3cacf84ed5

specifically this line
https://github.com/httprb/http/blob/3a06852cad71d8611d2a17fdfca8da3cacf84ed5/lib/http/response.rb#L37

But I think this is an issue on our test being too dependent on the class of the URI object.
So I think we should update this test to not depend on the class of the URI object,
and assert the original uri string is preserved.

I'm exploring more, and working on a fix.

presidentJFK:
The reason for the having to update http_rb_spec_helper.rb from

```ruby
:status     => response.status.to_s,
```
to
```ruby
:status     => response.status.code.to_s,
```

is because of this commit, wrapping the status in an object

https://github.com/httprb/http/commit/136e2ad37e857b260345f6c7e14a3f79e59dc901

so we could also update the code to just call code

```ruby
:status     => response.code.to_s,
```

PR #509, opened 2015-08-30
Spec fixes

presidentJFK:


presidentJFK:
Hey @bblimke do you want to take a look at this commit before merging.

I explored each of the failures, and included the relevant commits and info to why I made each change in the commit messages.

presidentJFK:
@bblimke Since this is only spec changes, and I feel like I have identified the reason for each, I am going to merge this in. This way I can start taking a look at the other issues with green specs.

bblimke:
@presidentJFK good call. Cheers! Sorry, I was away and haven't been able to look into it.

PR #511, opened 2015-09-02
Add rspec matcher descriptions

presidentJFK:


presidentJFK:
@arronmabrey alright everything looks good! Thanks for taking care of this :clap: 

I deleted that other branch, so the branch name would be the same as your original, and added one extra commit to unify the RequestExecutionVerifier spec language not to use should.

Issue #486, opened 2015-06-12
WebMock doesn't raise exception when specifying headers as symbols using Net::HTTP


steveklebanoff:
When making a Net::HTTP request specifying headers as symbols, an exception is raised.  If you stub the request using WebMock, an exception is not raised, and it acts as though the request is successful.

This can cause a major headache:  trying to make this request raises an exception in the real world, but passes when the request is stubbed with WebMock.

@cjlarose and I have created an isolated project to highlight this behavior.  Please see  https://github.com/cjlarose/webmock-net-http-headers-bug

bblimke:
@steveklebanoff looks like it's a problem in WebMock indeed.

It may not be trivial to fix it though. Net::HTTP doesn't handle invalid headers but an unexpected error is raised when expecting string. WebMock intercepts the request before it reaches problematic code in header.rb.

presidentJFK:
So I made a Pull Request #515 mirroring Net::HTTP's functionality, by just checking the headers for symbols before normalizing them and raising a NoMethodError with identical error message.

I need to see how the other http adapters work with headers, but this change at least fixes issue for Net::HTTP.

The only question is whether we should raise an error saying Net::HTTP doesn't except headers as symbols, which would be more useful than this error.

opinions? @bblimke @steveklebanoff 

otherwise I am going to merge this in, as it does prevent a false positive in a test.

cjlarose:
Nice work @presidentJFK! Left a [comment][1] on the PR.

[1]: https://github.com/bblimke/webmock/pull/515/files#r38891939

steveklebanoff:
Thank you for doing this @presidentJFK ! Looks awesome.

presidentJFK:
closing now that #515 is merged in.

bblimke:
well done @presidentJFK

akestner:
Run into some odd behavior related to this issue.  For some reason (with Ruby 2.3.0), my non-mocked requests with symbol headers work fine, but fail in my tests due to the associated PR (#515).  

Any ideas?

PR #515, opened 2015-09-06
Mirror Net::HTTP handling of headers as symbols

presidentJFK:
as brought up in #486

### DIFF
--- a/lib/webmock/http_lib_adapters/net_http.rb
+++ b/lib/webmock/http_lib_adapters/net_http.rb
@@ -273,8 +273,20 @@ module WebMock
 
       headers = Hash[*request.to_hash.map {|k,v| [k, v]}.inject([]) {|r,x| r + x}]
 
-      headers.reject! {|k,v| k =~ /[Aa]uthorization/ && v.first =~ /^Basic / } #we added it to url userinfo
+      # If you try and make a request with headers that are symbols,
+      # Net Http raises the following NoMethodError.
+      #
+      # WebMock normalizes headers when creating a RequestSignature,
+      # and will update all headers from symbols to strings.
+      # This means that you could have a test passing with WebMock, but failing in reality.
+      header_as_symbol = headers.keys.find {|header| header.is_a? Symbol}
+      if header_as_symbol
+        raise NoMethodError.new(
+          "undefined method `split' for :#{header_as_symbol}:Symbol (NoMethodError)`"
+        )
+      end
 
+      headers.reject! {|k,v| k =~ /[Aa]uthorization/ && v.first =~ /^Basic / } #we added it to url userinfo
 
       if request.body_stream
         body = request.body_stream.read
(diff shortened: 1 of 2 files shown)


lib/webmock/http_lib_adapters/net_http.rb
```diff
@@ -273,8 +273,20 @@ def self.request_signature_from_request(net_http, request, body = nil)
...
-      headers.reject! {|k,v| k =~ /[Aa]uthorization/ && v.first =~ /^Basic / } #we added it to url userinfo
+      # If you try and make a request with headers that are symbols,
+      # Net Http raises the following NoMethodError.
+      #
+      # WebMock normalizes headers when creating a RequestSignature,
+      # and will update all headers from symbols to strings.
+      # This means that you could have a test passing with WebMock, but failing in reality.
+      header_as_symbol = headers.keys.find {|header| header.is_a? Symbol}
+      if header_as_symbol
+        raise NoMethodError.new(
```
cjlarose: In response to [this comment][1]

> The only question is whether we should raise an error saying Net::HTTP doesn't except headers as symbols, which would be more useful than this error.

I think I would indeed prefer an exception making explicit that `Net::HTTP` doesn't accept symbol headers. Otherwise, if I'm a user of WebMock, I might get this exception and suspect WebMock of having a bug. It also doesn't help me much in actually fixing my code. The tradeoff, of course, is that WebMock wouldn't then be emulating `Net::HTTP` exactly. But I think situations where someone would be relying on this particular behavior being the same are incredibly rare.

[1]: https://github.com/bblimke/webmock/issues/486#issuecomment-138045396

steveklebanoff: The current error message does emulate Net::HTTP actual behavior but is a really confusing error message to get.  I definitely feel @cjlarose 's point about raising an error message that is more descriptive.  I could go either way though :+1:

steveklebanoff:
Thank you! I hope this saves some headaches.

presidentJFK:
@steveklebanoff @cjlarose I updated the code to raise an ArgumentError specifying Net::HTTP doesn't take headers as symbols.

So the only questions are:
  * is ArgumentError the right type of error?
  * and what should the error message be? should the invalid headers be included?

Any feedback/input would be greatly appreciated

There is also a problem assigning header as symbols in Ruby 1.8.7 which I skipped for now in the test suite. However I plan on exploring it soon.

steveklebanoff:
@presidentJFK `ArgumentError` feels like the right error class to me, and I think the error message you have (`Net:HTTP does not accept headers as symbols`) is fine -- I don't feel strongly about the invalid headers being included or not

presidentJFK:
So it's hard to assign headers as symbols using Ruby 1.8.7 and Net::HTTP, because Net::HTTP calls downcase on any headers passed in, and Symbols don't respond to downcase in 1.8.7.
But I have still updated the spec to test if someone somehow got into this weird state.

@steveklebanoff @cjlarose either of you want to take one final look before I merge this in?

steveklebanoff:
Looks good to me!  Thanks again

presidentJFK:
Great! Merging in.

PR #517, opened 2015-09-08
convert query values to strings before comparison

presidentJFK:


presidentJFK:
Some things I am still working on/figuring out:
  * figure out it this is the best place for this conversion
  * figure out if we should limit the classes we call .to_s on in the conversion
  * try this out with hash_including

presidentJFK:
@Fire-Dragon-DoL you want to take a look and see if this fixes your problem, or add any other feedback?

Fire-Dragon-DoL:
@presidentJFK I can't test against my code because it doesn't exist anymore (we changed it a lot), but by reviewing your PR it definitely looks like it fixes it entirely.

presidentJFK:
It looks like this doesn't solve your original problem with hash_including, so I am giving this another look.

Fire-Dragon-DoL:
Mhhh that's weird, handling the hash as string-values only on "both sides" of the hash_including, should work

presidentJFK:
I made a couple mistakes when trying to fix this.  I thought a simple one time stringify in the QueryMapper would fix the original issue.

But the problem with my original fix was it didn't take into account the HashIncludingMatcher. which handles its own comparision with #==, and was not converting the values to strings before comparing.

I have added the first spec you originally posted to the test suite, which exercises this exact functionality.
And updated the code to stringify before comparing values, fixing your test.

I now just have to clean up some changes I made, move specs to the right place, and double check
a couple more assumptions I have made, and we can get this merged in. And actually solve your original problem!

PR #524, opened 2015-09-24
Remove state from ManticoreAdapter

hobodave:
ManticoreAdapter was not thread-safe. Multiple threads would access
the same Client instance, which stored `@uri`, `@method`, and `@options`.
Multiple concurrent requests would cause data inconsistencies and
map responses to the incorrect request.

Fixing the thread-safety exposed several failing tests in the Manticore
specs. The gist of this is, Webmock allows you to stub with headers that
have Array values. e.g. `:headers => { 'Accept' => ['image/jpg', 'image/png'] }

However, you cannot pass an Array as a header value to the Manticore library.
It is the user's responsibility to concatenate this to a comma-separated string.

In order to match the configured stub, you have to naively split all
header values containing a comma into an Array, before attempting to match.

This feels wrong. However, I don't know a better way of doing it. My Webmock-fu
is weak.

### DIFF
--- a/lib/webmock/http_lib_adapters/manticore_adapter.rb
+++ b/lib/webmock/http_lib_adapters/manticore_adapter.rb
@@ -26,9 +26,6 @@ if defined?(Manticore)
 
         class WebMockManticoreClient < Manticore::Client
           def request(klass, url, options={}, &block)
-            @method = KLASS_TO_METHOD.fetch(klass)
-            @uri = url
-            @options = options
             super(klass, WebMock::Util::URI.normalize_uri(url).to_s, format_options(options))
           end
 
@@ -58,7 +55,20 @@ if defined?(Manticore)
           end
 
           def response_object_for(client, request, context, &block)
-            request_signature = generate_webmock_request_signature
+            if request.respond_to?(:entity) && !request.entity.nil?
+              body = Manticore::EntityConverter.new.read_entity(request.entity)
+            else
+              body = nil
+            end
+
+            headers = request.headers.each_with_object({}) do |(k, v), h|
+              h[k] = case v
+                     when /,/ then  v.split(',').map(&:strip)
+                     else v
+                     end
+            end
+
+            request_signature = generate_webmock_request_signature(request.method.downcase!, request.uri.to_s, body, headers)
             WebMock::RequestRegistry.instance.requested_signatures.put(request_signature)
 
             if webmock_response = registered_response_for(request_signature)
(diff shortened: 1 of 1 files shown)


hobodave:
@mikeknep Take a look?

hobodave:
Not sure why a Typhoeus spec failed on ruby-2.1.0 -- seems like a flake. Rebuild?

davidbegin:
you were right @hobodave Build is green now

lib/webmock/http_lib_adapters/manticore_adapter.rb
```diff
@@ -26,9 +26,6 @@ def self.disable!
 
         class WebMockManticoreClient < Manticore::Client
           def request(klass, url, options={}, &block)
-            @method = KLASS_TO_METHOD.fetch(klass)
```
mikeknep: With this line removed, we can also remove the `KLASS_TO_METHOD` constant

lib/webmock/http_lib_adapters/manticore_adapter.rb
```diff
@@ -58,7 +55,20 @@ def join_array_values(headers)
           end
 
           def response_object_for(client, request, context, &block)
-            request_signature = generate_webmock_request_signature
+            if request.respond_to?(:entity) && !request.entity.nil?
+              body = Manticore::EntityConverter.new.read_entity(request.entity)
+            else
+              body = nil
+            end
```
mikeknep: I'd prefer this pulled out into a separate method to keep `response_object_for` relatively high-level for readability. Also feels like there should be a better alternative to the `respond_to?` and `nil?` checks, but I'm not sure what off the top of my head.

lib/webmock/http_lib_adapters/manticore_adapter.rb
```diff
@@ -58,7 +55,20 @@ def join_array_values(headers)
...
+
+            headers = request.headers.each_with_object({}) do |(k, v), h|
+              h[k] = case v
+                     when /,/ then  v.split(',').map(&:strip)
+                     else v
+                     end
+            end
```
mikeknep: Sigh. So many header formatting things across the adapters in this project, and this is now the 2nd header-formatting method in this adapter alone...

mikeknep:
@hobodave great catch, and I love the intent here. Some nitpick comments above. I'd like to take a stab at standardizing header-formatting across adapters if possible... I may take a stab at it on the train back from StrangeLoop tomorrow.

mikeknep:
@davidbegin I think the CI failures are unrelated flakes again. I made some nitpick edits into @hobodave's fork, and this is now working in our real-world use of this.

davidbegin:
@mikeknep you are correct about the flakes, I retriggered the failing specs, and CI is now green.

I will take a look over these changes tonight and try and get this merged in.

davidbegin:
Thanks @hobodave @mikeknep! I rebased this on master and merged it in.

PR #538, opened 2015-10-28
prevents form adding an extra =true to urls

davidbegin:
Without this fix:
    * Requests to:
      http://www.google.com?query

    * Stub registered as:
      http://www.google.com?query=true


bblimke:
@davidbegin great that you found the fix :)

PR #541, opened 2015-11-04
Eliminate warnings(assigned but unused) and remove unnecessary to_s

rochefort:


lib/webmock/request_execution_verifier.rb
```diff
@@ -34,26 +34,26 @@ def does_not_match?
     end
 
     def description
-      "request #{request_pattern.to_s} #{quantity_phrase.strip}"
+      "request #{request_pattern} #{quantity_phrase.strip}"
     end
 
     def failure_message
-      failure_message_phrase(is_negated=false)
+      failure_message_phrase(_is_negated = false)
```
keram: Hi @rochefort, may ask  question, what you think it is purpose even this unused ``_is_negated`` variable here?
Wouldn't be better if ``is_negated`` was argument of ``failure_message``? Something like:
```ruby
def failure_message(is_negated = false)
  failure_message_phrase(is_negated)
end
```
which would fix the warning as well or by using keywords arguments:
```ruby
def failure_message
  failure_message_phrase(is_negated: false)
end
...

def failure_message_phrase(options = {})
  negation = options.fetch(:is_negated, false) ? "was not" : "was"
...
```

rochefort: hi @keram, thank you for pointing it out.

I wrote `_is_negated` for like that:
- keyword arguments are not using in this project
- it is better wrote `_is_negated`  is clearly

But come to think of it, `_is_negated` is strange as arguments.
I like style of keyword arguments better, but it is for ruby 2.0 later.
And webmock supports MRI 1.8.7 later.
So I will remove it from arguments.

davidbegin:
Since we support back to Ruby 1.8.7 right now, we can't have keyword arguments without writing ruby version specific code.

bblimke:
@rochefort thank you for these improvements!

PR #544, opened 2015-11-05
Add RequestStub helper methods for accessing request signatures

brettlangdon:
While working with `WebMock` we have found that we would like to be able to easily access the `WebMock::RequestSignature`s that correspond to a given `WebMock::RequestStub`.

The reason for this is so that we can have more generic stubs and then write our assertions in the tests against the requests that were made rather than building very specific stubs.

This PR adds in the following methods to `WebMock::RequestStub`:

- `requests` - filter the requests in `WebMock::RequestRegistry` which match `@request_pattern`
- `last_request` - the last request from `requests` (`requests[-1]`)
- `first_request` - the first request from `requests` (`requests[0]`)

This allows us to write tests similar to the following:

```ruby
let!(:stub) { WebMock::API.stub_request(:post, "elasticsearch.url/people/person/_search")
                          .to_return(File.new("response.raw")) }

before(:example) do
    visit("/endpoint")
end

it "properly queried elasticsearch" do
    request = stub.last_request
    data = JSON.load(request.body)
    expect(body["query"]).to(eq("match_all" => {}))
end
```

As we are still somewhat new to `WebMock` we are also open to suggestions on how we should be doing this.

### DIFF
--- a/lib/webmock/request_stub.rb
+++ b/lib/webmock/request_stub.rb
@@ -53,6 +53,28 @@ module WebMock
       end
     end
 
+    # Fetch all requests from `WebMock::RequestRegistry` which match this stubs `@request_pattern`
+    # @return [Array] the requested `WebMock::RequestSignature`s which match this stubs `@request_pattern`
+    def requests
+      request_registry = WebMock::RequestRegistry.instance
+      signatures = request_registry.requested_signatures.hash.keys()
+      return signatures.select { |signature|
+        @request_pattern.matches?(signature)
+      }
+    end
+
+    # Fetch the first request made for this stub
+    # @return [WebMock::RequestSignature|nil] the first request signature for this stub, or nil if none were made
+    def first_request
+      return requests[0]
+    end
+
+    # Fetch the last request made for this stub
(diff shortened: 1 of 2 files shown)


bblimke:
@brettlangdon thank's for submitting this pull request and for explaining the intentions clearly.

I can see it being useful, but I'm not sure it's the right way to achieve that yet.

RequestStub should not be coupled with or know anything about RequestRegistry. Perhaps there should be a method `requests_for_stub(stub)` available somewhere.

In your example there is only one stub, but there may be multiple stubs matching the request in various ways, and RequestRegistry doesn't tell you which one has been used to handle it.

Please note that RequestRegistry also stores 'real', unstubbed, requests, before the stub was added.

Of course there are not as likely scenarios as the one you presented.

brettlangdon:
@bblimke thanks for the reply.

Yeah, I can see how coupling them together would not be ideal.

I would be more than happy to head back to the drawing board for this.

davidbegin:
This is an interesting idea! I am excited to see new implementations. But for now I am closing this PR. Please open a new one anytime.

PR #546, opened 2015-11-10
hash_counter: avoid possible RuntimeErrors on JRuby

kyrylo:
I stumbled upon this exception while running my tests on JRuby 1.7:

```
RuntimeError:
       can't add a new key into hash during iteration
     # ./vendor/jruby/1.9/bundler/gems/webmock-4d40934217e8/lib/webmock/util/hash_counter.rb:15:in `put'
     # ./vendor/jruby/1.9/bundler/gems/webmock-4d40934217e8/lib/webmock/util/hash_counter.rb:14:in `put'
     # ./vendor/jruby/1.9/bundler/gems/webmock-4d40934217e8/lib/webmock/http_lib_adapters/net_http.rb:77:in `request'
     ...my_frames...
```

Even though we don't iterate anything here, this simple change was able
to fix my problems.

**EDIT:** don't mind the webmock version in my backtrace, I was just testing things and depended on my fork.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 1.22.4
+
+  * Fix the "can't add a new key into hash during iteration" message in hash_counter.rb in JRuby 1.7
+
+    Thanks to [kyrylo](https://github.com/kyrylo)
+
 ## 1.22.3
 
   * Return "effective_url" attribute in Typhoeus::Response
--- a/lib/webmock/util/hash_counter.rb
+++ b/lib/webmock/util/hash_counter.rb
@@ -14,8 +14,7 @@ module WebMock
         @lock.synchronize do
           val = hash[key]
           hash[key] = (val || 0) + num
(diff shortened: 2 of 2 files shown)


kyrylo:
Sorry, this shouldn't be merged as it doesn't fix anything. I believe it's a thread-safety issue with someone's code. Could be my fault.

kyrylo:
I tried to debug this very thoroughly, but unfortunately I don't understand the reason of that error. Even if it's a thread issue, we have a lock there. It makes no sense for me.

I tried to hammertest the HashCounter class (#put and #each were invoked in two separate threads), but couldn't reproduce that error.

bblimke:
@kyrylo ok, please give a shout if you manage to create example that can be reproduced easily. 

The change you made doesn't make any difference indeed. 

If you can show it's a problem with webmock please open an issue.

I'm going to close this pull request. New one will have to be created if needed.

kyrylo:
Sure, thank you!

PR #547, opened 2015-11-11
Add support for JSONClient (a subclass of HTTPClient)

nepalez:


nepalez:
There is something odd in travis specs today - seems like they fail randomly :(

davidbegin:
Thanks for submitting this! 

And you are correct about the travis specs, there are some intermittent failures that I haven't been able to reproduce locally. But haven't given them a lot time yet.

I'll try and take a look at your PR soon, so we can merged in.

lib/webmock/http_lib_adapters/httpclient_adapter.rb
```diff
@@ -105,6 +103,44 @@ def build_httpclient_response(webmock_response, stream = false, req_header = nil
...
   end
 
+  class WebMockHTTPClient < HTTPClient
+    alias_method :do_get_block_without_webmock, :do_get_block
+    alias_method :do_get_stream_without_webmock, :do_get_stream
+
+    include WebMockHTTPClients
+  end
+
+  if defined? ::JSONClient
```
bblimke: It seems there is duplication here. HTTPClientAdapter with enable and disable methods has already been defined and now it's being redefined.

Is there a reason to define `HTTPClientAdapter` twice in case JSONClient is defined? Perhaps the JSONClient related code can be wrapped in `if defined? ::JSONClient`?

nepalez: Yeah, you're right.
Updated

nepalez:
@bblimke could you please look at this PR once again? :)

bblimke:
@nepalez thanks for reminder :) everything's fine. cheers!

nepalez:
Thank you :)

PR #551, opened 2015-12-08
Clarify RSpec support

deivid-rodriguez:
I've been using webmock with RSpec 3.x and it seems to work fine. Maybe it's better to just say "RSpec is supported", just like with Minitest and Test::Unit.


deivid-rodriguez:
Build is failing because the `addressable` gem dropped Ruby 1.8 support in its latest version. You probably want to add some conditional dependencies or drop 1.8 support as well.

bblimke:
Cheers!

deivid-rodriguez:
Thanks!

PR #555, opened 2015-12-22
Allow curb 0.9.0

matthewrudy:
So I went through the code,
and the tests pass fine with 0.9.0.

Trying to understand the issue I tried 0.8.7
and indeed it freezes the tests.

But 0.8.8 passes too.

So it seems that this was a unique issue with 0.8.7.
Perhaps we need to modify the `VersionChecker` to have a blacklist of known bad versions?

### DIFF
--- a/lib/webmock/http_lib_adapters/curb_adapter.rb
+++ b/lib/webmock/http_lib_adapters/curb_adapter.rb
@@ -5,7 +5,7 @@ rescue LoadError
 end
 
 if defined?(Curl)
-  WebMock::VersionChecker.new('Curb', Curl::CURB_VERSION, '0.7.16', '0.8.7').check_version!
+  WebMock::VersionChecker.new('Curb', Curl::CURB_VERSION, '0.7.16', '0.9.0').check_version!
 
   module WebMock
     module HttpLibAdapters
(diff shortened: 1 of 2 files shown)


davidbegin:
Interesting, thanks for looking into this! 

I am not sure why this restriction was in place, however maybe @bblimke has more insight.

But no matter what I think we should figure out what is different about 0.8.7 and figure what's the best way to blacklist a version. I will look into both shortly.

Also I am pretty sure the CI is failing because master CI is currently failing. I am currently working on that, and once its done we can rebase this to make sure its truly green on CI.

matthewrudy:
@davidbegin thanks.

OT: Travis is breaking because addressable 2.4.0 was released a few weeks ago
that no longer supports 1.8.6

matthewrudy:
@davidbegin to add to the Feedback on curb 0.8.7.

I tried it,
and it literally froze in a spec.

But I didn't stick around to try and debug it.

maartenvanvliet:
:+1:

bblimke:
I have just released version 1.24.0 which accepts curb 0.9.x versions.
It shows warning is Curb version 0.8.7 is used.

PR #556, opened 2015-12-22
Ruby 1.8.7 requires addressable < 2.4.0

matthewrudy:
Fixes travis build for 1.8.7

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,3 +16,7 @@ end
 platforms :jruby do
   gem 'jruby-openssl'
 end
+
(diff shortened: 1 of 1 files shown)


davidbegin:
Thanks, taking a look now!

davidbegin:
It looks like this chang also affected JRuby, which causes the specs to fail.

So I am experimenting over here https://github.com/bblimke/webmock/pull/557 to only restrict addressable for 1.8

matthewrudy:
@davidbegin sure.

PR #557, opened 2015-12-22
Fix 1.8 and make ci pass

davidbegin:


davidbegin:
I still need to figure out what has changed to cause this spec to start failing in the first place.
I tried to create the exact dependencies from a passing Travis CI build, but the spec was still failing.

But now I have a better idea of where the change must be coming from, RSpec.

matthewrudy:
@davidbegin no worries.
Thanks for your time.

I'll leave these issues in your hands.

It works fine in our app running curb 0.9.0 and Ruby 2.2.2
But just get some devs worrying about the warning,
so thought it was worth exploring.

Thanks for your help.

PR #559, opened 2015-12-27
Update http adapter and fix ci

davidbegin:


davidbegin:
The specs that are failing are because of this commit in http rb https://github.com/httprb/http/commit/19aad3121ba1bd2b0ac30558c4fc3a79b3ccba07

davidbegin:
@ixti we just need to figure out what the best way to handle this change: https://github.com/httprb/http/commit/19aad3121ba1bd2b0ac30558c4fc3a79b3ccba07 in WebMock.

My last commit adds a quick hack

```ruby
def perform(request, options)
  return __perform__(request, options) unless webmock_enabled?
  WebMockPerform.new(request) { __perform__(request, options) }.exec

rescue HTTP::Error => e
  raise Errno::ECONNREFUSED.new(e.message)
end
```

this seems to fix the specs, but I'm not sure if it is best approach. If you have any insight or suggestions that would be greatly appreciated, if not I will explore a more permanent fix soon.

ixti:
Your fix looks about right. Although I guess it's better to utilize `#cause` of exceptions and backfill it (for ruby < 2.1):

``` ruby

unless StandardError.new.respond_to? :cause
class HTTP::Error
  attr_reader :cause
  def initialize(msg, cause = $!)
    @cause = cause
    super msg
  end
end

# ...

def perform(request, options)
  return __perform__(request, options) unless webmock_enabled?
  WebMockPerform.new(request) { __perform__(request, options) }.exec

rescue HTTP::Error => e
  e = e.cause if e.cause
  raise e 
end
```

ixti:
I have changed my mind. Why do you handle that exception in first place? You should not.

ixti:
Oh. I got why you have added it there.... Well As I said - no need to do that in `HTTP::Client` in fact :D

ixti:
I will update my PR

JuanitoFatas:
@davidbegin Cut a new release for this? :blush:

davidbegin:
@JuanitoFatas already on it! https://github.com/bblimke/webmock/pull/564

JuanitoFatas:
Thanks for prompt reply. : )

PR #581, opened 2016-02-03
Generic `:allow` descriptor

pablobm:
I have a situation where I would like Webmock to allow connections, except to **a number** of sites.

I could do that with a regular expression, but it can become quite unwieldy, and more as more sites are added to the list. Instead, I thought it would be better if I can pass a `call`-able (eg: Proc or lambda) as the `:allow` in `disable_net_connect!`. For example:

```Ruby
WebMock.disable_net_connect!(:allow => ->(url){ url.to_str.include?('httpstat') })
```

Does this make sense? Is this desiderable?

While implementing this, I noticed that there aren't good specs to cover the different formats accepted for `allow`, so I added that, then added a case for my proposal, as above.

The resulting spec (actually, shared example: `spec/acceptance/shared/allowing_and_disabling_net_connect.rb
`) works well, but it is a far cry from DRY. I was thinking the facility to say whether a URL is allowed or not could be factored out and unit-tested separately. Then the acceptance tests could rely on a stubbed true/false response from this separate component.

If you think this new feature is acceptable, I could provide such refactoring in the future. (Unfortunately not just now; I have duties to attend).

### DIFF
--- a/README.md
+++ b/README.md
@@ -462,22 +462,67 @@ Net::HTTP.get('www.something.com', '/')    # ===> Failure
 Net::HTTP.get('localhost:9887', '/')    # ===> Allowed. Perhaps to Selenium?
 ```
 
-### External requests can be disabled while allowing any hostname or port or parts thereof
+### External requests can be disabled while allowing specific requests
+
+Allowed requests can be specified in a number of ways.
+
+With a `String` specifying a host name:
 
 ```ruby
-WebMock.disable_net_connect!(:allow => "www.example.org:8080")
+WebMock.disable_net_connect!(:allow => 'www.example.org')
 
 RestClient.get('www.something.com', '/')    # ===> Failure
+RestClient.get('www.example.org', '/')      # ===> Allowed
+RestClient.get('www.example.org:8080', '/') # ===> Allowed
+```
 
-RestClient.get('www.example.org', '/')    # ===> Failure.
+With a `String` specifying a host name and a port:
 
-RestClient.get('www.example.org:8080', '/')    # ===> Allowed
+```ruby
+WebMock.disable_net_connect!(:allow => 'www.example.org:8080')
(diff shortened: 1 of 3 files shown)


bblimke:
@pablobm thanks for the pull request and for improving the specs :)

"Does this make sense? Is this desiderable?" - It definitely makes sense.
I never had a reason to do that, and nobody ever requested that feature, but possibly it is desirable too.

Would you mind sharing your usecase and explaining why would one extend the list of allowed hosts dynamically?
Do you use this feature for tests?

pablobm:
Sure thing, this is the use case.

This is part of a large refactoring where an app uses several APIs, and we are replacing them one by one. One of these APIs doesn't work 100% yet and we are providing fixtures for some features. Another one is going to disappear completely, and we are mocking it to return 404 until properly removed. At times, we want the app to reach for the real APIs, configured with envvars. This is where Webmock comes into play.

Using envvars, we tell the app which APIs to mock and which not to. When a mock is activated, it's added to a list that ultimately is fed to `disable_net_connect!` (using the feature in this PR). Anything not activated should be accessible normally via the Internet.

Webmock is therefore used in development and staging mode. It's also used in test, but independently from this use case.

I'm thinking I probably should add something to the README too?

bblimke:
Thanks for sharing your case. I'm always happy to hear about WebMock being
used for different scenarios than just tests. I suspected it must be use-case for development rather than testing.

I'm happy to merge it. It would be good to add at least one example to README to make people aware of this feature.

pablobm:
README updated :-)

bblimke:
Excellent. Thanks for improving the README section :)

bblimke:
@pablobm it's now released as 1.23.0

pablobm:
Yay! Thank you very much.

PR #582, opened 2016-02-08
Allow for webmock response to have body chunks

cedricpim:
Recently I was using [vcr](https://github.com/vcr/vcr) along with [google-api-client](https://github.com/google/google-api-ruby-client) and had some issues with `vcr` recording the body of the responses. After some investigation, I realized that a request was done using `Hurley` (relies on `HTTPClient` and sends a block as well to get the chunks, however `webmock` was only calling that block but not taking care of storing the body of the response which resulted in an empty body.

### DIFF
--- a/lib/webmock/http_lib_adapters/httpclient_adapter.rb
+++ b/lib/webmock/http_lib_adapters/httpclient_adapter.rb
@@ -72,13 +72,19 @@ if defined?(::HTTPClient)
 
         res = if stream
           do_get_stream_without_webmock(req, proxy, conn, &block)
+        elsif block
+          body = ''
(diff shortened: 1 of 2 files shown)


bblimke:
@cedricpim thank you for this fix!

fotinakis:
Excellent! This will fix my problems with VCR + google-api-client as well (https://github.com/google/google-api-ruby-client/issues/356#issue-133458453).

@bblimke, when do you think the next WebMock release will be?

bblimke:
It's now released as 1.23.0

fotinakis:
:+1: Thanks!

gondalez:
solved my problem too, thanks! 👍

PR #586, opened 2016-02-19
HTTPClient: Fix manually specifying Authorization header

mkarnebeek:
## Problem

When manually specifying the Authorization header like
```Ruby
c = HTTPClient.new
c.request :get, uri.omit(:userinfo, :query), query, body, {'Authorization' => 'Basic #{digest}'}
```
instead of like:
```Ruby
c = HTTPClient.new
c.set_basic_auth(nil, 'username', 'password')
c.request :get, uri.omit(:userinfo, :query), query, body, headers
```
WebMock does not catch the username and password and does not put it in the uri. (not the missing username and password in the url)
```
WebMock::NetConnectNotAllowedError: Real HTTP connections are disabled. Unregistered request: GET http://localhost/some/endpoint
```
I would expect (note the username and password being present in the url):
```
WebMock::NetConnectNotAllowedError: Real HTTP connections are disabled. Unregistered request: GET http://username:password@localhost/some/endpoint
```

## Why don't i just use `set_basic_auth()`?

`set_basic_auth()` is not thread-safe, and specifying it in the URI is also not supported (from what I read from the code and tried). So, the only way for me to specify the credentials in a thread-safe way is to specify the header myself.

## But, HTTPClient advertises to be thread-safe!

Yes they do, and yes they are, but setting credentials is not.

When using `HTTPClient::IncludeClient`'s `include_http_client()` (singleton behaviour for HTTPClient), it serves the same instance of HTTPClient to every thread. This causes `HTTPClient`'s instance variable `@www_auth` to be an singleton, which causes the `WWWAuth`'s instance variable `@basic_auth` to be a singleton, which causes `BasicAuth`'s instance variable `@cred` to be a singleton. 

`BasicAuth`'s instance variable `@cred` being a singleton causes the usernames and passwords to mix up in the following example

```Ruby
class FunkyClient
  extend HTTPClient::IncludeClient
  include_http_client(:method_name => :my_client)

  def do_authenticated_call(method, username, password, uri, query = nil, body = nil)
    my_client.set_basic_auth(uri, username, password) #this sets the 'singleton' `@cred`
    my_client.request method, uri, query, body # this risks executing with @cred from another thread
  end
end

# say, the following two lines are executed in their own threads, simultaniously
FunkyClient.new.do_authenticated_call(:get, 'user1', 'pass1', URI('http://localhost/')
FunkyClient.new.do_authenticated_call(:get, 'user2', 'pass2', URI('http://localhost/')
```

I tried solving this in HTTPClient, but as far as I can see, this might actually be unsolvable, because of how authentication is handled (binding credentials to an url, instead of to a session). This makes it impossible to do multi-threading/keepalive with different credentials to the same url.

## The step back to WebMock

This makes `set_basic_auth()` unsafe for multi-threading and unusable in my use-case. So i set the authentication header manually, which causes WebMock to not recognise the credentials. 

Do exactly the same in Net::HTTP and it works as expected, so i decided to fix WebMock.

## Solution

Have WebMock's `httpclient_adapter.rb` support reading the credentials directly from the header, instead of relying on HTTPClient -> WWWAuth -> BasicAuth having credentials set.

This pull request makes the changed needed to do that, plus it fixes the clearing of that header when credentials are found.

### DIFF
--- a/lib/webmock/http_lib_adapters/httpclient_adapter.rb
+++ b/lib/webmock/http_lib_adapters/httpclient_adapter.rb
@@ -172,9 +172,6 @@ if defined?(::HTTPClient)
     uri.port = req.header.request_uri.port
     uri = uri.omit(:userinfo)
 
-    auth = www_auth.basic_auth
-    auth.challenge(req.header.request_uri, nil)
-
     @request_filter.each do |filter|
       filter.filter_request(req)
     end
@@ -186,34 +183,9 @@ if defined?(::HTTPClient)
     end
     headers = headers_from_session(uri).merge(headers)
 
-    auth_cred = nil
-    auth_cred = auth.get(req) if auth.scheme == 'Basic'
-    headers.each do |k,v|
-      if k =~ /[Aa]uthorization/
-        if v.is_a? Array
-          v.each do |v|
-            auth_cred ||= v if v =~ /^Basic /
-          end
-        elsif v.is_a? String
-          auth_cred ||= v if v =~ /^Basic /
-        end
-      end
-    end
-
-    if auth_cred
-      userinfo = WebMock::Util::Headers.decode_userinfo_from_header(auth_cred)
-      userinfo = WebMock::Util::URI.encode_unsafe_chars_in_userinfo(userinfo)
-      headers.reject! do |k, v|
-        if k =~ /[Aa]uthorization/
-          if v.is_a? Array
-            v.reject! { |v| v =~ /^Basic / }
-            v.length == 0
-          elsif v.is_a? String
-            v =~ /^Basic /
-          end
-        end
-      end
-      uri.userinfo = userinfo
+    if auth_cred = auth_cred_from_www_auth(req) || auth_cred_from_headers(headers)
+      remove_authorization_header headers
+      uri.userinfo = userinfo_from_auth_cred auth_cred
     end
 
     signature = WebMock::RequestSignature.new(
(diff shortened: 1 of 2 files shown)


lib/webmock/http_lib_adapters/httpclient_adapter.rb
```diff
@@ -186,10 +186,15 @@ def build_request_signature(req, reuse_existing = false)
     end
     headers = headers_from_session(uri).merge(headers)
 
-    if (auth_cred = auth.get(req)) && auth.scheme == 'Basic'
+    if ((auth_cred = auth.get(req)) && auth.scheme == 'Basic') || auth_cred = headers.select {|k,v| k =~ /[Aa]uthorization/ && v.select { |v2| v2 =~ /^Basic / }.first }.values.first.first
```
mkarnebeek: Possibly the `.values.first.first` can be solved in a nicer way, but i was a bit lazy. It might also break when `headers = {"Authorization": []}`

lib/webmock/http_lib_adapters/httpclient_adapter.rb
```diff
@@ -186,10 +186,15 @@ def build_request_signature(req, reuse_existing = false)
...
       userinfo = WebMock::Util::Headers.decode_userinfo_from_header(auth_cred)
       userinfo = WebMock::Util::URI.encode_unsafe_chars_in_userinfo(userinfo)
-      headers.reject! {|k,v| k =~ /[Aa]uthorization/ && v =~ /^Basic / } #we added it to url userinfo
```
mkarnebeek: This was actually broken, because v is an Array.

bblimke:
@mkarnebeek thank you for this very well described pull request :)

The logic makes sense and I'll be happy to merge it.

In order to merge it this pull request requires acceptance spec example covering this scenario in `httpclient_spec.rb`

It also makes sense to do some refactoring and perhaps extract this code to a separate method/methods, since `build_request_signature` now became very long and complex (if it wasn't already).

mkarnebeek:
Added spec and extracted authentication code from build_request_signature into separate methods.

bblimke:
Nice :+1:

mkarnebeek:
When do you expect it to show up on RubyGems?

bblimke:
`Pushed webmock 1.24.1 to rubygems.org.`

mkarnebeek:
Speedy! :D

PR #590, opened 2016-02-28
Improve parsing of params on request

cedricpim:
I went through the issues and saw nothing related to this. When stubbing requests, I was having difficult when dealing with a more complex structure (array of hashes).

Here the approach followed is iterate through the params and create a new hash on the array if the key already exists in the last hash.


bblimke:
Thank you. The query mapper is quite complex and I hope this change won't break other scenarios,
but the change you made makes sense so it's worth giving it a try.

cedricpim:
No problem. I tried to change little as possible, just let me know if you run into any issues :+1:

bblimke:
released as 1.24.2

PR #592, opened 2016-03-09
Make the Typhoeus adapter also read the Authorization header for user info

knu:
When using Typhoeus via Faraday, the :userpwd option is never used because Faraday is a backend independent framework in which Faraday::Request::BasicAuthentication handles basic authentication simply by setting the Authorization header without bothering with any backend specific option.

### DIFF
--- a/lib/webmock/http_lib_adapters/typhoeus_hydra_adapter.rb
+++ b/lib/webmock/http_lib_adapters/typhoeus_hydra_adapter.rb
@@ -53,8 +53,16 @@ if defined?(Typhoeus)
         def self.build_request_signature(req)
           uri = WebMock::Util::URI.heuristic_parse(req.url)
           uri.path = uri.normalized_path.gsub("[^:]//","/")
-          if req.options[:userpwd]
-            uri.user, uri.password = req.options[:userpwd].split(':')
+
+          headers = req.options[:headers]
+          if authorization = req.options[:userpwd] ||
+                             headers.find { |name, value|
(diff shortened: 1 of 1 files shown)


bblimke:
@knu thanks for the pull request.

The code requires refactoring (extracting to methods) before it could be merged and some test to verify that behaviour.

I'm concerned about removing the authorization header. In case the request stub expected the header to be present. Any thoughts?

knu:
Hi, thanks for taking a look.

Yes, this is just a bug report with a minimal patch and really needs some test.

As for the header removal, other adapters are doing this so I just followed that:
- https://github.com/bblimke/webmock/blob/master/lib/webmock/http_lib_adapters/net_http.rb#L290
- https://github.com/bblimke/webmock/blob/master/lib/webmock/http_lib_adapters/httpclient_adapter.rb#L211

The original problem is here, that you have to change how to mock a request with basic auth dependent on the backend library:
https://github.com/cantino/huginn/pull/1340/files#diff-502af1f8d307aaf7af861f16d2688b86R1040

knu:
Perhaps the Authorization header should not be removed in all adapters in the first place because it makes it impossible to mock a request specifying the Authorization header.

bblimke:
@knu yes, and that's the change in WebMock 2.0. https://github.com/bblimke/webmock/blob/2.0/CHANGELOG.md 
it's currently released as 2.0.0.beta1

knu:
Cool!  I'd really appreciate your decision to make the change taking advantage of the timing of starting off a new major version series.  Thank you!

PR #597, opened 2016-03-13
Only raises symbol headers exception if `RUBY_VERSION` < 2.3.0

### LINKED ISSUE #595: WebMock shouldn't raise exception when specifying headers as symbols
As of Ruby `v2.3.0` it's acceptable to specify `Net::HTTP` headers as symbols (https://github.com/ruby/ruby/commit/1a98f56ae14724611fc8f7c220e470d27f6b57e4), but webmock currently raises `ArgumentError.new("Net:HTTP does not accept headers as symbols")` if you do.

This was introduced in #515, I'd propose wrapping that behavior with a check of `RUBY_VERSION`.

akestner:
Fixes #595.

### DIFF
--- a/lib/webmock/http_lib_adapters/net_http.rb
+++ b/lib/webmock/http_lib_adapters/net_http.rb
@@ -304,7 +304,8 @@ module WebMock
     end
 
     def self.validate_headers(headers)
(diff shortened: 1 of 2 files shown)


bblimke:
Thank you! Looks good.

PR #598, opened 2016-03-18
Don't shadow outer variable in reject! block.

Manfred:
This emits a warning when running a test suite with warnings enabled. Renaming the outer variables to name and value make it slightly more readable and removes the warning.


bblimke:
Thank you! More readable code is always welcome :)

PR #601, opened 2016-03-30
make easier to run non-network specs quickly.

timdiggins:
spec_helper does potentially expensive check every time rspec starts
(waits for non responsive server 173.194.113.176, in theory only 5s, but in my experience a lot longer)
but if you want to repeatedly run a single non-network (e.g. unit) spec this is tiresome. In this modificatiion it will skip this expensive check if the env var NO_CONNECTION is already set.


bblimke:
Good idea!

bblimke:
I changed network connectivity detection to be much faster.

PR #605, opened 2016-04-13
document automatic enabling

### LINKED ISSUE #583: webmock enabled on require
Requiring Webmock immediately enables it.  I believe that this is bad behavior and should change.  Given that it's probably expected behavior at this point, it should at least be documented. Probably something in 
https://github.com/bblimke/webmock#disabling-and-enabling-webmock-or-only-some-http-client-adapters

Offending commit: https://github.com/bblimke/webmock/commit/b3d60450e49e724a38a40d4f98fafdfed611ebd3#diff-9c446f800dcca622179ac9fd536cb7e1R100

lanej:
* Fixes #583

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,10 @@ require 'webmock'
 include WebMock::API
 ```
 
+### Automatically enabled
+
+`require 'webmock'` loads the library AND enables `WebMock`.  Add `WebMock.disable!` after loading the gem to disable this behavior.
+
 ## Examples
 
 


lanej:
@bblimke wasn't sure how much is really necessary to say about this.  Might be worth mentioning if you're going to change the behavior in the future.

bblimke:
@lanej thank you! good idea to put that in the docs.

PR #611, opened 2016-04-28
WebMock should load "base64"

taiki45:
WebMock uses base64 but it doe not load base64 by itself. It may cause `NameError: uninitialized constant WebMock::Util::Headers::Base64`.


bblimke:
Thanks!

PR #612, opened 2016-04-28
Resolve RSpec warning

taiki45:
RSpec warns:

WARNING: Using `expect { }.not_to raise_error(SpecificErrorClass,
message)` risks false positives, since literally any other error
would cause the expectation to pass, including those raised by Ruby
(e.g. NoMethodError, NameError and ArgumentError), meaning the code you
are intending to test may not even get reached. Instead consider using
`expect {}.not_to raise_error`. This message can be supressed by setting:
`RSpec::Expectations.configuration.warn_about_potential_false_positives = false`


bblimke:
and thank you again :)

taiki45:
:)

PR #613, opened 2016-04-28
Resolve warning about `Object#timeout`

taiki45:
`Object#timeout` is deprecated.


bblimke:
Thank you!

taiki45:
Thank you for quick response!

PR #615, opened 2016-05-03
Fix regex comparison

jlhonora:
Before this change:

```
> WebMock.net_connect_explicit_allowed?(/.*example.*/, 'http://example.com')
0
```

After:

```
> WebMock.net_connect_explicit_allowed?(/.*example.*/, 'http://example.com')
true
```

I was having trouble with some non-shareable case, this change fixes it.


bblimke:
Looks good. Thank you!

PR #616, opened 2016-05-05
Add config notice for codeclimate

softwaregravy:
Adding webmock to a project using codeclimate breaks the codeclimate reporter: https://github.com/codeclimate/ruby-test-reporter/issues/23

Add a line in the docs to give the solution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,14 @@ include WebMock::API
 WebMock.enable!
 ```
 
+### Using Codeclimate?
+
+ Add this to your helper file (wherever you added the above):
+ 
+ ```ruby
+ WebMock.disable_net_connect!(allow: 'codeclimate.com')
+ ```
+
 ## Examples
 
 


bblimke:
Thank for this.

My only concern is whether it's the right direction. I.e there are known issues with travis gem and possibly many other gems as well, since WebMock blocks connections and I'm not sure adding them to README is the right place. 

Shall we expect people to search for "codeclimate" in the README?

Any thought?

softwaregravy:
My initial thoughts are 1) right now, when people hit this error and start googling, they end up reading through a bunch of issues on github. Finding the solution takes time and is non-trivial to find (not saying it's hard, just not 30 seconds). IMHO, the goal of documentation is to make things easier to use and save people time, which this has the potential to do. 2) are there lots of these issues? Are they crowding out the readme? if so, maybe create a page in the wiki or something and direct people there from the readme.

PR #619, opened 2016-05-18
Fix warnings

cupakromer:
This is a collection of minor updates to remove Ruby warnings from the gem. These should all be very low risk changes. By fixing these warnings it helps to eliminate noise in projects which use the gem and have warnings enabled. It's always kind for gems to produce zero warnings if possible.

The most visible to projects that use `webmock` are the warnings coming from the `lib` directory. While I was running the specs I noticed that several additional warnings came from some spec support files. I went ahead and cleaned those up as well.

I left off the local changes I made to the `spec/spec_helper.rb` which expose the warnings when running the specs as that's more a decision the core team needs to make. Also, JRuby has some gems like `manticore` and `cookiejar` which produce a lot of warnings making those builds very noisy. I'm happy to add them back and re-submit the PR if you'd like.

### DIFF
--- a/lib/webmock/request_body_diff.rb
+++ b/lib/webmock/request_body_diff.rb
@@ -15,9 +15,10 @@ module WebMock
       HashDiff.diff(request_signature_body_hash, request_stub_body_hash)
     end
 
-    private
-
     attr_reader :request_signature, :request_stub
+    private :request_signature, :request_stub
+
+    private
 
     def request_signature_diffable?
       request_signature.json_headers? && request_signature_parseable_json?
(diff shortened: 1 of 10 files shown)


bblimke:
Thank you! 👍

bblimke:
Fix released in version 2.0.3

cupakromer:
Awesome thanks! 😸

lib/webmock/request_body_diff.rb
```diff
@@ -15,9 +15,10 @@ def body_diff
       HashDiff.diff(request_signature_body_hash, request_stub_body_hash)
     end
 
-    private
-
     attr_reader :request_signature, :request_stub
+    private :request_signature, :request_stub
+
+    private
```
pda: Declaring `attr_reader` after `private` is [no longer a warning](https://bugs.ruby-lang.org/issues/10967) as of Ruby 2.3, so this is a legacy workaround. Retaining it may or may not be worthwhile to prevent warnings in old versions of ruby.

Issue #621, opened 2016-05-23
Curb do not trigger on_debug call back


ravenesi:
Hi,
web mock curb implementation do not trigger on_debug callback ... because of that i did this work in initializers... 

spec/support/webmock_curb_fix.rb

``` ruby
# frozen_string_literal: true
if defined?(Curl) && defined?(WebMock)
  module Curl
    WebMockCurlEasy.class_eval do
      def on_debug
        yield if block_given?
      end
    end
  end
end
```

bblimke:
@ravenesi any idea why it doesn't work with webmock?

ravenesi:
@bblimke indeed our WebMock implementation do not implement proper on_debug callback other callbacks are there i can create PR is is it needed :)

bblimke:
a PR would be very welcome :)

PR #627, opened 2016-06-01
Fix curb on debug

pavel-jurasek-bcgdv-com:
This PR fixing issue: https://github.com/bblimke/webmock/issues/621

### DIFF
--- a/lib/webmock/http_lib_adapters/curb_adapter.rb
+++ b/lib/webmock/http_lib_adapters/curb_adapter.rb
@@ -77,8 +77,6 @@ if defined?(Curl)
         else
           raise WebMock::NetConnectNotAllowedError.new(request_signature)
         end
-
-        build_webmock_response if defined?( @on_debug )
       end
 
       def build_request_signature
--- a/lib/webmock/http_lib_adapters/curb_adapter.rb
+++ b/lib/webmock/http_lib_adapters/curb_adapter.rb
@@ -77,6 +77,8 @@ if defined?(Curl)
         else
           raise WebMock::NetConnectNotAllowedError.new(request_signature)
         end
+
+        #build_webmock_response if defined?( @on_debug )
       end
 
       def build_request_signature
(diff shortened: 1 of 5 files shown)


pavel-jurasek-bcgdv-com:
https://travis-ci.org/pavel-jurasek-bcgdv-com/webmock should work

bblimke:
@pavel-jurasek-bcgdv-com thanks for that.

I'm wondering if "faking" on_debug actually makes sense.
This code just pretends to be a real Curl.

If someone wants to debug what's going on, I'm not sure seeing a data generated by WebMock,
is what user wants to see.

What would be the usecase of using on_debug with WebMock enabled?

pavel-jurasek-bcgdv-com:
@bblimke well i add on_debug statements just because simplecov do not hit then all code.

Use case:
Building abstract business client (internally using curb). If this client will contain debug option (Foo.client.debug = true) and within the class will be specified:

```ruby
@curb.on_debug do |foo, bar|
  >> loging to NewRelict in my case
end
```
 test coverage will be not 100% since on_debug is not implemented in WebMock. It will produce 100% coverage without WebMock. I can simplify on_debug statements but i tried to produce as close as possible results.

bblimke:
I see. Thanks for explaining.

What you implemented, is it exactly what curb would output every time?

I wouldn't want to mislead people who use on_debug to trace some issue, with incorrect data.

pavel-jurasek-bcgdv-com:
@bblimke as close as i can go it not handle follow location requests but for intention (e.g. on_debug handling) should be fine.

lib/webmock/http_lib_adapters/curb_adapter.rb
```diff
@@ -77,6 +77,8 @@ def curb_or_webmock
         else
           raise WebMock::NetConnectNotAllowedError.new(request_signature)
         end
+
+        build_webmock_response if defined?( @on_debug )
```
bblimke: why is that needed? isn't webmock response already built?

pavel-jurasek-bcgdv-com:
@bblimke removed :) thanks to check code ! i over looked this.

bblimke:
Excellent :)

bblimke:
released as 2.1.0

PR #628, opened 2016-06-14
Typo Authorization

weppos:
Just a small typo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,7 +218,7 @@ Net::HTTP.start('www.example.com') do |http|
 end    # ===> Success
 ```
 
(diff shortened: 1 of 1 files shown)


bblimke:
@weppos cheers!

I wondered if there is anything named "Athoriation". Interesting is that if you search for "Athoriation" on google, webmock comes second ;)

weppos:
I was actually upgrading from 1.x to 2.1 and investigating some failures caused by the use of the authorization in the URL. 😉

weppos:
Closing as it was fixed at some point.

PR #632, opened 2016-06-26
reset postion

nakajijapan:
Error occured when FILE IO's postion is END.
I think that you should reset file's position when webmock builds a request.


bblimke:
Can you please include the spec that demonstrates why these 3 lines are needed? Without a filing spec there is no reason for them to be there :)

nakajijapan:
ok!

PR #625, opened 2016-06-01
Add support for patch

pavel-jurasek-bcgdv-com:
This PR preparing WebMock to use this PR in CURB core: 
https://github.com/taf2/curb/pull/301

PR #635, opened 2016-07-17
Add suport to match body on PATCH for Curb

lucasgomide:


lucasgomide:
The build broke, look at #637

bblimke:
This should be compatible with taf2/master branch.

Since it's not released yet and it's possible it's going to be changed again,
I'm thinking about reverting support for Curb.patch completely,
and just keep it on branch (or onpull request), until Curb with patch support
is officially released.

lucasgomide:
I'm not understand when you say: `until Curb with patch support is officially released.`. 
Because it's possible make request using PATCH on Curb.

I'll remake this to be compatible with taf2/master branch. But before need revert #625 

What do you think ?

bblimke:
I thought it's only possible to do patch requests with Curb from master.

If specs are going to pass against latest Curb 9.3, then it's fine.

I'd like to avoid having dependencies on development versions from github,
since these are likely to change again.

lucasgomide:
Great. You are right !

lucasgomide:
@bblimke i'll close it, because on #639 has the revert necessary to work using the last release of curb

PR #652, opened 2016-10-08
several methods for httpclient_adapter defined on main level

bpieck:
I moved them into the module where they really are used in. I tried tests (still green) and tried with an own project, where we are using rack-oauth2 gem (which uses httpclient). Tests were still running as expected. Since I did not add new functionality, I did not write any tests.


bblimke:
Looks good. Thank you!

PR #655, opened 2016-10-24
Fix flaky HTTPClient adapter spec

lautis:
If the clock moves to next second during the spec run, date header might be different. Mock time to ensures that date headers of request signatures match.


bblimke:
Thanks. 

I wonder if the time matters in this example, since it's not clear where it comes from. Perhaps the assertion should be improved.

lautis:
HTTPClient sets the time header: https://github.com/nahi/httpclient/blob/f7b36d2257407dfbb807ab004c144b5a991ac5b1/lib/httpclient/session.rb#L733.

Changing the assertion would do the trick, but there's more work as everything else but one field in the RequestSignature should match. An alternative would be to set Date header in the requests explicitly.

bblimke:
Ok, makes sense!

PR #657, opened 2016-10-28
add refute_requested

grosser:
@bblimke @irwaters

### DIFF
--- a/test/shared_test.rb
+++ b/test/shared_test.rb
@@ -22,7 +22,6 @@ module SharedTest
   end
 
   def test_error_on_non_stubbed_request
-    default_ruby_headers = (RUBY_VERSION >= "1.9.1") ? "{'Accept'=>'*/*', 'User-Agent'=>'Ruby'}" : "{'Accept'=>'*/*'}"
     assert_raise_with_message(WebMock::NetConnectNotAllowedError, %r{Real HTTP connections are disabled. Unregistered request: GET http://www.example.net/ with headers}) do
       http_request(:get, "http://www.example.net/")
     end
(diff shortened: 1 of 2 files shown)


test/shared_test.rb
```diff
@@ -22,7 +22,6 @@ def test_assert_not_requested_with_stub_and_block_raises_error
   end
 
   def test_error_on_non_stubbed_request
-    default_ruby_headers = (RUBY_VERSION >= "1.9.1") ? "{'Accept'=>'*/*', 'User-Agent'=>'Ruby'}" : "{'Accept'=>'*/*'}"
```
grosser: was not used

test/shared_test.rb
```diff
@@ -48,7 +47,7 @@ def test_verification_that_expected_request_didnt_occur
     end
   end
 
-  def test_verification_that_expected_request_didnt_occur
+  def test_verification_that_expected_stub_didnt_occur
```
grosser: had the same name as the test above ... so never got executed

bblimke:
@grosser thanks. is `refute` a popular term when it comes to asserting? Do any test frameworks use it?  Perhaps it is. I'm asking because I've never seen `refute` in any test DSL.

grosser:
Minitest ships with it and imo it's the nicer alternative to assert_not which reads weird and does not align

https://github.com/seattlerb/minitest/search?utf8=%E2%9C%93&q=refute

and using it a bunch here:
https://github.com/zendesk/samson/search?utf8=%E2%9C%93&q=refute

bblimke:
@grosser excellent. Thanks!

PR #659, opened 2016-11-04
Correct Net::HTTP open timeout exception class

gabetax:
Net::HTTP raises a `Net::OpenTimeout`, a subclass of `Timeout::Error`, on open timeouts. It used a different class in 1.9 vs 2.x: https://github.com/ruby/ruby/commit/09f27873ed653bd4afb5538b669e2362f3b1bf6e

This is relevant because libraries like Faraday implement specific behavior on `Net::OpenTimeout`: https://github.com/lostisland/faraday/blob/v0.9.2/lib/faraday/adapter/net_http.rb#L28 and it can cause significantly different error handling in production vs under webmock.

This PR could may cause consumer's tests to break if they are asserting on the timeout class by name instead of hierarchy, but it implements the actual behavior of the library being stubbed.

### DIFF
--- a/lib/webmock/http_lib_adapters/net_http.rb
+++ b/lib/webmock/http_lib_adapters/net_http.rb
@@ -168,13 +168,28 @@ module WebMock
 
           response.extend Net::WebMockHTTPResponse
 
-          raise Timeout::Error, "execution expired" if webmock_response.should_timeout
+          if webmock_response.should_timeout
+            raise timeout_exception, "execution expired"
+          end
 
           webmock_response.raise_error_if_any
 
           yield response if block_given?
 
           response
+        end
+
+        def timeout_exception
+          if defined?(Net::OpenTimeout)
+            # Ruby 2.x
+            Net::OpenTimeout
+          elsif defined?(Net::HTTP::OpenTimeout)
+            # Ruby 1.9
+            Net::HTTP::OpenTimeout
+          else
+            # Fallback, if things change
+            Timeout::Error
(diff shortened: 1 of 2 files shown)


bblimke:
@gabetax awesome! thank you.

dentarg:
Hmm, not sure this timeout business is as clear-cut as one would think.

Why not make `to_timeout` raise `Net::ReadTimeout`? For that, Faraday will raise `Faraday::TimeoutError`, which is what we expected in our specs (so they indeed broke with this change).

This change makes Faraday raise `Faraday::ConnectionFailed`, which covers a lot more cases than just a timeout: https://github.com/lostisland/faraday/blob/v0.10.1/lib/faraday/adapter/net_http.rb#L13-L25

But it's a bit confusing to me why Faraday choose to (explicitly) handle one timeout exception (`Net::ReadTimeout`) as `ConnectionFailed` and another (implicitly) (`Net::ReadTimeout`) as `TimeoutError`. They could both be `TimeoutError` because [both inherit from Ruby's `Timeout::Error`](https://github.com/ruby/ruby/blob/v2_4_0/lib/net/protocol.rb#L72-L78).

@gabetax @bblimke

gabetax:
@exviva The exception class raised by ruby changed. If you stub your tests to force the old exception, it may make your tests green, but it's not going to behave correctly in production. I recommend paling an custom exception list containing `Faraday::ConnectionFailed` when building in the Faraday middleware.

PR #663, opened 2016-12-03
Support Ruby 2.4

koic:
This PR made the following changes in preparation Ruby 2.4.

- Fix deprecated warnings caused by [unifies Fixnum and Bignum into Integer](https://bugs.ruby-lang.org/issues/12005)
- Fix failure test caused by [add keyword arguments for Net::BufferedIO#initial attributes](https://github.com/ruby/ruby/commit/dad2382270b8bcfc56a0e2dd73c6201e63ed1a7c#diff-00a99d8c71daaf5fc60a050da41f7261L82)

### DIFF
--- a/lib/webmock/http_lib_adapters/net_http.rb
+++ b/lib/webmock/http_lib_adapters/net_http.rb
@@ -248,6 +248,9 @@ class StubSocket #:nodoc:
     @closed ||= true
   end
 
+  def close
+  end
+
   def readuntil(*args)
   end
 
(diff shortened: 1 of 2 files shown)


bblimke:
@koic thank you!

bblimke:
Actually I was too fast merging it.

This breaks Ruby 1.9.3 support, which doesn't support named arguments.

koic:
Thanks for your replying.

I think that drop Ruby 1.9.3 support is good idea, because 2 reasons below.

1. [Ruby 1.9.3 was EOL](https://www.ruby-lang.org/en/news/2015/02/23/support-for-ruby-1-9-3-has-ended)
2. Some gems have already finished supporting Ruby 1.9.3

I propose to finish support of Ruby 1.9.

How about this?

bblimke:
Good idea!

koic:
Thank you for dropped Ruby 1.9.3. However, the codes of supporting Ruby 2.4 was still reverted. Is this intentional?

bblimke:
I forgot to merge revert of revert. I'll yank 2.3.0 an release 2.3.1 tonight

bblimke:
2.3.1 is now out. Thank you!

koic:
Thanks too!

PR #669, opened 2016-12-10
Use 1.9 hash key syntax in stub suggestions

tarmo:


tarmo:
The build failure appears to be unrelated to my change as it also happens on master both when I run it locally and also @ https://travis-ci.org/tarmo/webmock/builds/182789108

bblimke:
@tarmo nice, thank you!

PR #670, opened 2016-12-10
Update Travis CI configuration

ojiry:
* Use latest ruby version
* Remove duplicated minor version


tarmo:
FYI the reason the spec for "HTTPClient when a client instance is re-used for another identical request invokes the global_stub_request hook for each request" fails is because webmock is using latest version of httpclient which just released a new version. The exact change in httpclient causing this spec to fail is @ https://github.com/nahi/httpclient/commit/c28d1d4bb391c4e988537cd2079d4025e3e15b0d#diff-2f54812a2786a9c04529e1b7236542ebR299  The spec calls verify_mode=VERIFY_NONE twice but the second call does not trigger change_notify because the client thinks that verify_mode didn't change, yet change_notify is what is needed to make the spec pass again. I'm however not sure why exactly that is the case and if it's a bug in httpclient or in how webmock is using it.

ojiry:
@tarmo Thank you! I will investigate the problem 🙏

ojiry:
I closed this PR because it cannot be resolved immediately.

PR #671, opened 2016-12-12
Add option to hide all registered request stubs

shaunmcgeever:


bblimke:
@shaunmcgeever thank you for this pull request.

I never had issues with displaying registered stubs, therefore I'm interested,
to why you would like to prevent them being displayed.

Could you please convince me, why this feature is worth adding? :)

Issue #606, opened 2016-04-20
RuntimeError - can't add a new key into hash during iteration


burgalon:
```
RuntimeError - can't add a new key into hash during iteration
    /Users/burg/.rvm/gems/ruby-2.3.0/gems/webmock-1.22.3/lib/webmock/util/hash_counter.rb:15:in `block in put'
    /Users/burg/.rvm/gems/ruby-2.3.0/gems/webmock-1.22.3/lib/webmock/util/hash_counter.rb:14:in `synchronize'
    /Users/burg/.rvm/gems/ruby-2.3.0/gems/webmock-1.22.3/lib/webmock/util/hash_counter.rb:14:in `put'
    /Users/burg/.rvm/gems/ruby-2.3.0/gems/webmock-1.22.3/lib/webmock/http_lib_adapters/net_http.rb:77:in `request'
    /Users/burg/.rvm/rubies/ruby-2.3.0/lib/ruby/2.3.0/net/http.rb:1378:in `send_request'
    /Users/burg/git/infinidat/infinibox-webgui2/webgui/test/spec/support/sinatra_proxy.rb:26:in `block in <class:SinatraProxy>'
```

The above is being caused by RequestRegistry.times_executed which iterates on the hash while some other thread might be setting it.

Working on a pull request

grosscr:
Is there a fix coming for this? I've run into this issue as well.

PR #674, opened 2016-12-28
Synchronous hash counter access from request registry.

starsirius:
Fixes https://github.com/bblimke/webmock/issues/606

We had the same
```
can't add a new key into hash during iteration
```
issue in our specs, and I think the [issue](https://github.com/bblimke/webmock/issues/606) above did capture the cause.

My naive try here seemed to fix the problem for us, but as I don't fully understand the details, this may not be a good solution. Still thought I would PR it and get feedback. Happy to iterate on this and add some specs.

### DIFF
--- a/lib/webmock/request_registry.rb
+++ b/lib/webmock/request_registry.rb
@@ -14,11 +14,9 @@ module WebMock
     end
 
     def times_executed(request_pattern)
-      self.requested_signatures.lock.synchronize do
-        self.requested_signatures.hash.select { |request_signature, times_executed|
-          request_pattern.matches?(request_signature)
-        }.inject(0) {|sum, (_, times_executed)| sum + times_executed }
-      end
+      self.requested_signatures.select do |request_signature|
+        request_pattern.matches?(request_signature)
+      end.inject(0) { |sum, (_, times_executed)| sum + times_executed }
     end
 
     def to_s
--- a/lib/webmock/request_registry.rb
+++ b/lib/webmock/request_registry.rb
@@ -14,9 +14,11 @@ module WebMock
     end
 
     def times_executed(request_pattern)
-      self.requested_signatures.hash.select { |request_signature, times_executed|
-        request_pattern.matches?(request_signature)
-      }.inject(0) {|sum, (_, times_executed)| sum + times_executed }
+      self.requested_signatures.lock.synchronize do
+        self.requested_signatures.hash.select { |request_signature, times_executed|
+          request_pattern.matches?(request_signature)
+        }.inject(0) {|sum, (_, times_executed)| sum + times_executed }
+      end
     end
 
     def to_s
--- a/lib/webmock/util/hash_counter.rb
+++ b/lib/webmock/util/hash_counter.rb
@@ -4,6 +4,7 @@ module WebMock
   module Util
     class Util::HashCounter
       attr_accessor :hash
+      attr_accessor :lock
       def initialize
         self.hash = {}
         @order = {}
--- a/lib/webmock/util/hash_counter.rb
+++ b/lib/webmock/util/hash_counter.rb
@@ -4,7 +4,6 @@ module WebMock
   module Util
     class Util::HashCounter
       attr_accessor :hash
-      attr_accessor :lock
       def initialize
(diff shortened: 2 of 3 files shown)


lib/webmock/request_registry.rb
```diff
@@ -14,9 +14,11 @@ def reset!
     end
 
     def times_executed(request_pattern)
-      self.requested_signatures.hash.select { |request_signature, times_executed|
-        request_pattern.matches?(request_signature)
-      }.inject(0) {|sum, (_, times_executed)| sum + times_executed }
+      self.requested_signatures.lock.synchronize do
```
starsirius: Use the same mutex to synchronize between threads.

bblimke:
@starsirius Thank you!

The solution looks fine to me.

Are you able to provide a deterministic spec to reproduce the problem?

It would be best to hide the mutex logic inside the Hashcounter,
and not expose that to request_registry.

Perhaps instead of `.hash.select` there should be a safe `select` method defined on Hashcounter,
or some kind of safe Enumerator exposed.

starsirius:
Thanks for the comments, @bblimke! I will find some time add a spec and better encapsulate the mutex logic in Hashcounter.

starsirius:
Great! The spec [failed](https://travis-ci.org/bblimke/webmock/builds/210801282) with expected `can't add a new key into hash during iteration` runtime error on Travis. Going to work on the fixes next!

screen shot 2017-03-13 at 10 24 02 pm

spec/unit/request_registry_spec.rb
```diff
@@ -43,6 +43,25 @@
...
     end
+
+    describe "multithreading" do
+      let(:request_pattern) { WebMock::RequestPattern.new(:get, "www.example.com") }
+
+      # Reproduce a multithreading issue that causes a RuntimeError:
+      #   can't add a new key into hash during iteration.
+      it "works normally iterating on the requested signature hash while another thread is setting it" do
+        thread_injected = false
+        allow(request_pattern).to receive(:matches?).and_wrap_original do |m, *args|
```
starsirius: The idea of the spec here is that when `request_pattern.matches?` was called, we inject a second thread that writes to the same internal hash. It gives a deterministic results but it's coupled with the `times_executed` implementation. :man_shrugging:

spec/unit/request_registry_spec.rb
```diff
@@ -43,6 +43,25 @@
...
+          unless thread_injected
+            thread_injected = true
+            Thread.new { WebMock::RequestRegistry.instance.requested_signatures.put(:abc) }.join(0.1)
```
starsirius: Had to `join` immediately to produce deterministic failure in the current implementation. However, it will also cause a deadlock in the proposed solution (since both `put` and `select` are waiting for the other to release the lock). Setting a limit by `join(0.1)` here addressed both cases.

spec/unit/request_registry_spec.rb
```diff
@@ -43,6 +43,25 @@
...
+          end
+          m.call(*args)
+        end
+        expect(WebMock::RequestRegistry.instance.times_executed(request_pattern)).to eq(2)
+        sleep 0.1 while !WebMock::RequestRegistry.instance.requested_signatures.hash.key?(:abc)
```
starsirius: Make sure the second thread did `put` an additional key.

lib/webmock/util/hash_counter.rb
```diff
@@ -22,6 +22,13 @@ def get key
         end
       end
 
+      def select(&block)
+        return unless block_given?
+        @lock.synchronize do
+          hash.select &block
```
starsirius: Simply proxy the `select` to `Hash#select` implementation.

starsirius:
@bblimke Apologies for the delay. I updated the PR with your comments. Let me know how you think about this or want me to clarify anything. Thanks!

bblimke:
@starsirius Thank you! Well done. I can see it has been very tricky to reproduce it.

starsirius:
Hi @bblimke Looks like we have some good stuff for a new version. Do you think it's safe to release v2.3.3 soon? Thanks!

PR #675, opened 2016-12-30
Add at_least matchers for fakeweb-style expectations

mltsy:
Don't know why these at_least matchers weren't included originally, but it seems like they should be there... :)


bblimke:
@mltsy I don't see any reason why there matchers have not been yet added either :)

Thanks!

mltsy:
Thank you :smile:  (and Happy New Year!) :tada:

PR #680, opened 2017-02-03
Http.rb 2.2 fix

krzyzak:
So, it turns out that Http.rb changed constructor for the `Http::Response::Body` class, which now requires 2, instead of one parameter.

This change broke WebMock, resulting that when you’re trying to call stubbed request, you’ll get following error:

```
require 'http'
HTTP::VERSION # => 2.2.0
HTTP.get("https://google.com") # => #<HTTP::Response/1.1 302 Found {"Cache-Control"=>"private", "Content-Type"=>"text/html; charset=UTF-8", "Location"=>"https://www.google.pl/?gfe_rd=cr&ei=75mUWLDZJMLi8Af7_6HYCA", "Content-Length"=>"259", "Date"=>"Fri, 03 Feb 2017 14:55:43 GMT", "Alt-Svc"=>"quic=\":443\"; ma=2592000; v=\"35,34\"", "Connection"=>"close"}>

require "webmock/rspec"
include WebMock::API
stub_request(:get, "https://google.com/").to_return(body: "hello!")
WebMock::VERSION #=>  => 2.3.2
HTTP.get("https://google.com")
ArgumentError: wrong number of arguments (given 1, expected 2..3)
```

This pull request bumps http.rb dependency to `>= 2.2.0` *and* fixes that behaviour.

### DIFF
--- a/lib/webmock/http_lib_adapters/http_rb/response.rb
+++ b/lib/webmock/http_lib_adapters/http_rb/response.rb
@@ -14,7 +14,7 @@ module HTTP
       def from_webmock(webmock_response, request_signature = nil)
         status  = Status.new(webmock_response.status.first)
         headers = webmock_response.headers || {}
-        body    = Body.new Streamer.new webmock_response.body
+        body    = Body.new(webmock_response.body, Streamer.new(webmock_response.body))
         uri     = normalize_uri(request_signature && request_signature.uri)
 
         return new(status, "1.1", headers, body, uri) if HTTP::VERSION < "1.0.0"
--- a/webmock.gemspec
+++ b/webmock.gemspec
@@ -27,7 +27,7 @@ Gem::Specification.new do |s|
     s.add_development_dependency 'typhoeus', '>= 0.5.0'
   end
 
-  s.add_development_dependency 'http',            '>= 0.8.0'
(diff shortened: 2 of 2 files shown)


matthewrudy:
@krzyzak I saw this too.

Not sure if this is the right fix,
but the problem was introduced in this PR on the http gem
https://github.com/httprb/http/pull/375/files#diff-05df387cd9e9cd3b09c4a966e7510519

bblimke:
@krzyzak thank you!

I'm not sure this is the right fix either, since building against version 2.2.1 of HTTP, released yesterday, broke the build again.

krzyzak:
High five guys, as I have no idea if that’s the right fix as well :D 

I’m not really active in terms of what happens under the hood neither in Webmock, nor in HTTP.rb (although I really like both projects).

I checked, and yeah, it looks like it was regression in HTTP.rb, rather than bug in webmock, and it looks like they restored compatibility with webmock in 2.2.1

fcheung:
Is this obsolete? Both the referenced versions of http.rb and WebMock are long gone, and WebMock seems to work just fine with current version of http.rb

bblimke:
@fcheung thanks. let me close it and if anyone finds it needs to be reopened, please do that.

PR #687, opened 2017-02-16
Fix invalid UTF-8 when unencoding uri

### LINKED ISSUE #679: [v1] ArgumentError: invalid byte sequence in UTF-8
This bug only seems to be present for the webmock prior to v2 and Ruby versions >= 2.1.0.

When stubbing a request with the following features:

- stubbed request uses a regular expression for URI
- actual request has basic auth
- actual request uses non-standard port (e.g. not 443 for HTTPS)

A minimal reproduction can be seen at: https://gist.github.com/english/0fcdaa9be05c88e35feb4ed8197ba07e, but here's a paraphrased version:

```rb
stub_http_request(:get, /foo/).to_return(body: "abc")
Net::HTTP.start("example.com", 9000, use_ssl: true) do |http|
  request = Net::HTTP::Get.new('/foo')
  request['Authorization'] = 'Basic foo'
  http.request(request)
end
```

When running the above with webmock v1.24.6 (and at least v1.21.0), and Ruby 2.1 or greater results in a `ArgumentError: invalid byte sequence in UTF-8` exception.

I'm unsure as to whether the v1 branch is currently maintained for bug fixes? If not, this bug report may at least be useful for others searching with the same issue.

english:
Fixes #679

Most of the URIs that come through
`WebMock::Util::URI.variations_of_uri_as_strings` end up getting forced
`ASCII_8BIT` encoding, except for URIs that neither have a `/` path,
standard port mapping nor a http scheme.

This caters for that specific code path by forcing the encoding of the
URI string returned from `Addressable::URI.unencode`.

### DIFF
--- a/lib/webmock/util/uri.rb
+++ b/lib/webmock/util/uri.rb
@@ -87,7 +87,9 @@ module WebMock
 
       def self.uris_encoded_and_unencoded(uris)
         uris.map do |uri|
-          [ uri.to_s, Addressable::URI.unencode(uri, String).freeze ]
+          unencoded = Addressable::URI.unencode(uri, String)
+          unencoded.force_encoding(Encoding::ASCII_8BIT) if unencoded.respond_to?(:force_encoding)
+          [ uri.to_s, unencoded.freeze ]
         end.flatten
       end
 
--- a/spec/unit/util/uri_spec.rb
+++ b/spec/unit/util/uri_spec.rb
@@ -115,6 +115,16 @@ describe WebMock::Util::URI do
       end
     end
 
+    it "should find all variations of uris with https, basic auth, a non-standard port and a path" do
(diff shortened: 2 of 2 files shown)


bblimke:
@english why is it on branch 1.x?

english:
The bug described in https://github.com/bblimke/webmock/issues/679 was only present in the 1.x release. You mentioned in https://github.com/bblimke/webmock/issues/679#issuecomment-277126211 that you would consider releasing a patch version if someone were to submit a fix.

The last 1.x release (1.24.6) matches the 1.x branch so I had assumed you would like the PR submitted against that branch. Is that correct?

bblimke:
@english right :)

it looks like all specs are failing. not related to your changes, but perhaps it's worth limiting dependencies.

Issue #688, opened 2017-02-22
Suggested stub body should be valid JSON when request header contains 'Accept'=>'application/json'


redbar0n:
When the request has a `'Accept'=>'application/json'` header set, then Webmock suggests a `:body => ""`:

    You can stub this request with the following snippet:

    stub_request(:post, <some_url>).
          with(:body => <some_body>,
               :headers => {'Accept'=>'application/json' }).
          to_return(:status => 200, :body => "", :headers => {})

but doing a `JSON.parse(response.body)` will give an:

`JSON::ParserError: A JSON text must at least contain two octets!`

What Webmock should suggest instead is simply a `:body => "{}"`

    You can stub this request with the following snippet:

    stub_request(:post, <some_url>).
          with(:body => <some_body>,
               :headers => {'Accept'=>'application/json' }).
          to_return(:status => 200, :body => "{}", :headers => {})

Since that would be parseable by Ruby's JSON library:
```
> irb
>> JSON.parse("")
JSON::ParserError: 743: unexpected token at ''
	from /Users/Magne/.rvm/gems/ruby-2.3.1/gems/json-2.0.3/lib/json/common.rb:156:in `parse'
	from /Users/Magne/.rvm/gems/ruby-2.3.1/gems/json-2.0.3/lib/json/common.rb:156:in `parse'
	from (irb):3
	from /Users/Magne/.rvm/rubies/ruby-2.3.1/bin/irb:11:in `<main>'
>> JSON.parse("{}")
=> {}
```

bblimke:
@edbar0n I guess that very much depends on what the mythical server would really respond with.
That server could take into account Accept headers or ignore them.

`to_return(:status => 200, :body => "", :headers => {})` is just a simplest response,
to give people idea how to declare a stub, but implementing the actual response
is something WebMock can't make any assumptions about.
It's a programmers job to declare the specific response.

redbar0n:
I realize that the Webmock suggestion is just a starting point for implementing the actual response. But doesn't it make sense that the starting point which Webmock suggests should be a good starting point, i.e. something that would work?

It's an assumption either way: That the server would ignore the Accept header (like it currently does), or that it takes it into account (responding with JSON if the client asks for JSON). I'd argue that it would be more common to take it into account, and that it would give a better starting point for programmers. Who would implement a request that asks for JSON, and not initially expect to actually receive JSON back in the response?

[...]

PR #691, opened 2017-03-06
Fix: Suggested stub body to_return should be valid JSON when request header contains 'Accept'=>'application/json'. It now suggests a return body with '{}'.

redbar0n:


redbar0n:
RE: https://github.com/bblimke/webmock/issues/688

redbar0n:
All tests ran green locally. Not sure why the Travis CI build failed.

› bundle exec rake test
Loaded suite /Users/Magne/.rvm/gems/ruby-2.3.1/gems/rake-12.0.0/lib/rake/rake_test_loader
Started
............

Finished in 0.030351 seconds.
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
12 tests, 14 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
395.37 tests/s, 461.27 assertions/s

bblimke:
@redbar0n `bundle exec rake`

`rake test` only runs test unit tests.

redbar0n:
Ok. Anything I need to do?

bblimke:
@redbar0n the code is broken as tests show, so it needs to be fixed before merging :)

Then I'd fix the duplication I mentioned it the comment, unless you have a different opinion?

redbar0n:
Ok. I'm not sure how to fix Travis CI test failures. 

Duplication? I can't find the comment you're referring to.

bblimke:
https://github.com/bblimke/webmock/pull/691#discussion_r105141327

redbar0n:
Done. :)

bblimke:
@redbar0n sorry, perhaps I wasn't clear.

When I asked 'is the first & necessary?', I referred to semantics not to the syntax.
Why do you need to check if `request_pattern` is present?

I don't think you have addressed my comment regarding duplication yet :)

redbar0n:
I see. I don't know if request_pattern can ever be nil, so I just check just in case. But if it cannot then that check can just be removed.

Was you comment regarding duplication the #691 (comment) link you gave? I can't access it for some reason. It just takes me to this same page.

bblimke:
You have to click on `Show outdated` to see the comment.

If there was a need for request_pattern checking then specs would fail if it was missing.

redbar0n:
Where do I find `Show outdated` on github? I can't see it anywhere on this page.

oliakremmyda:
Since the tests are passing now can we merge this PR?

oliakremmyda:
Also, I think it would be a good idea to squash all these commits into a single one :)

bblimke:
@oliakremmyda we can, indeed 👍 thanks for reminding about this. feel free to merge it.

PR #692, opened 2017-03-06
Ensure the correct read_timeout value is used

ckoenig:
The implementation of WebMockNetBufferedIO is ignoring the read_timeout values passed during creation. With Ruby MRI 2.3.3 [1] this works, since the value is updated later on, but Ruby MRI 2.4.0 [2] is not doing that anymore.

[1] https://github.com/ruby/ruby/blob/v2_3_3/lib/net/http.rb#L904
[2] https://github.com/ruby/ruby/blob/v2_4_0/lib/net/http.rb#L955

### DIFF
--- a/lib/webmock/http_lib_adapters/net_http.rb
+++ b/lib/webmock/http_lib_adapters/net_http.rb
@@ -257,7 +257,7 @@ module Net  #:nodoc: all
 
   class WebMockNetBufferedIO < BufferedIO
     def initialize(io, read_timeout: 60, continue_timeout: nil, debug_output: nil)
-      @read_timeout = 60
+      @read_timeout = read_timeout
       @rbuf = ''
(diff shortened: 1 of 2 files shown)


ckoenig:
I'm a bit confused by the failed build for jruby-9.0.5.0 since it seems to be working on my local machine.

bblimke:
Thank you!

dentarg:
@ckoenig thanks for fixing this :-)

@bblimke is it possible to get a release with this fix? anything I could do to help with that?

bblimke:
@dentarg it's now released as version 3.0.0

dentarg:
@bblimke thanks! really appreciate it, webmock is awesome

bblimke:
@dentarg cheers! :)

PR #697, opened 2017-04-12
Suppress `warning: `&' interpreted as argument prefix`

koic:
Here's a fix for a Ruby warning.

The following is a sample of reproduction.

```sh
% ruby -we 'def foo(&block); proc &block; end'
-e:1: warning: `&' interpreted as argument prefix
```

Thanks.


singhprd:
You beat me to it! Was getting the same thing with Ruby 2.4.1 Thanks @koic

bblimke:
Thank you!

bblimke:
released as 3.0.1

PR #698, opened 2017-04-15
Add hash excluding matcher to query option

Warrior109:


bblimke:
@Warrior109 thank you.

I believe that the actual `hash_excluding` method implementation is missing.

There are also specs missing that would verify that `hash_excluding` behaviour,
which would show that it's missing.

It's also worth updating the README

Warrior109:
I added hash_excluding_matcher_spec.rb with test hash excluding matcher. And i updated a Readme.md.

bblimke:
@Warrior109 you have added matcher and the spec for the matcher but you have not integrated hash_including into the webmock api and not tested it (in specs or tests)
Please see where hash_including is referenced in webmock https://github.com/bblimke/webmock/search?utf8=%E2%9C%93&q=hash_including&type=

Warrior109:
What about this version? All it`s ok? Or, maybe, i must do something again?

bblimke:
Almost there :)

One example in each of these 3 specs.

`test/shared_test.rb`
`spec/acceptance/shared/stubbing_requests.rb`
`spec/acceptance/shared/request_expectations.rb`

Looking at api_spec.rb examples I suggest using a hash that contains some some values and doesn't contain other values, and asserting a match instead of just copy and paste and changing `to` to `not_to`

Warrior109:
What about this version?)

bblimke:
Looking good now 👍

Warrior109:
Thanks a lot.)

bblimke:
Released as version 3.1.0. I'm sorry for the delay releasing it.

PR #707, opened 2017-06-28
Update homepage in the gemspec to be HTTPS

amatsuda:
The "homepage" in the gemspec (and so in rubygems.org page) points this repo via http protocol, but it redirects to https when we actually visit the page.


bblimke:
@amatsuda thank you!

PR #712, opened 2017-08-03
Patch raise_timeout_error for HTTP.rb adapter to return proper error

RickCSong:
HTTP.rb should raise a HTTP::TimeoutError rather than a HTTP::ConnectionError
when raising a timeout error.


bblimke:
@RickCSong thank you for the pull request!
@zanker thank you for confirming the correct fix!

zanker:
Thanks for the speedy review!

bblimke:
Released as version 3.1.0. I'm sorry for the delay.

PR #717, opened 2017-09-14
HTTPClient doesn't yield block on empty body

nurse:
By following code to reproduce, it doesn't call block.

Server:
```
#!/usr/bin/env ruby
require 'socket'
TCPServer.open("", 0) do |serv|
  addr = serv.addr
  puts "port: #{addr[1]}"
  while true
    Thread.start(serv.accept) do |s|
      print(s, " is accepted\n")
      p s.readpartial 900
      s.write "HTTP/1.1 200 OK\nContent-Type: text/html\n\n"
      s.close
    end
  end
end
```

Client:
```
require'httpclient'
HTTPClient.new.get("http://localhost:53503"){|r,c|p [r,c]}
```

### DIFF
--- a/lib/webmock/http_lib_adapters/httpclient_adapter.rb
+++ b/lib/webmock/http_lib_adapters/httpclient_adapter.rb
@@ -75,8 +75,10 @@ if defined?(::HTTPClient)
         elsif block
           body = ''
           do_get_block_without_webmock(req, proxy, conn) do |http_res, chunk|
-            body += chunk
-            block.call(http_res, chunk)
+            if chunk && chunk.bytesize > 0
+              body += chunk
+              block.call(http_res, chunk)
+            end
           end
         else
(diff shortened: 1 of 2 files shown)


bblimke:
@nurse thank you for the pull requests.

Have all specs passed in your environment? You can see that builds failed on Travis on all Ruby versions, but possibly this is not related to the pull request.

nurse:
@bblimke It seems because of typhoeus 1.3.0's incompatibility.
The failure also reproduces on trunk.

nurse:
Created a PR to fix that https://github.com/bblimke/webmock/pull/718

bblimke:
Thank you. I'm going to release it after I sort out the issue if two builds on travis failing.

bblimke:
Released as version 3.1.0

PR #718, opened 2017-09-14
Support typhoeus 1.3.0

nurse:
Typhoeus::Response::Header becomes delegated class instead of Hash
subclass from 1.3.0.


bblimke:
Awesome. Thank you!

bblimke:
Released as version 3.1.0

PR #721, opened 2017-10-02
Make it compatible with the newest http.rb version (3.0.0)

Bonias:
http.rb 3.0.0 have introduced HTTP::Request::Body abstraction (https://github.com/httprb/http/pull/412) - let's make webmock compatible with it.


bblimke:
Excellent. Thank you!

bblimke:
Released as version 3.1.0

PR #725, opened 2017-10-19
excon adapter: new defaults to empty hash

timogoebel:
Excon Adapter should default to an empty hash to allow `Excon::Connection.new`.

See:
https://github.com/excon/excon/blob/c92f240e0e90fc01beac4c52bbdbe2ed06b55263/lib/excon/connection.rb#L52


bblimke:
Yes, that makes sense. Thank you!

Issue #449, opened 2015-01-30
Response body can't be a hash


Morred:
When I stub a request and specify a hash for the response body, it gives me the following error:
```
WebMock::Response::InvalidBody:
    must be one of: [Proc, IO, Pathname, String, Array]. 'Hash' given
```
Is there a particular reason why it can't be a hash?

bblimke:
Please have a look at https://github.com/bblimke/webmock/pull/427

If response body is declared as a hash, it's not clear what should the hash be encoded to. JSON, XML, or perhaps url params? Therefore WebMock expects you to stringify hash to expected encoding. I.e {a: b}.to_json

Morred:
(Sorry for the late response, and thanks for your reply.)

I see, that makes sense. Maybe this should be mentioned in the Readme?

joshuaswilcox:
I am not sure that makes sense.  If a server is actually returning JSON as the body content, the code parsing can assume its a Hash content type, therefore calling JSON.parse(response) will fail if its already a Hash.  So its seems to me that Webmock should be able to send that as the response body, especially when a content-type of 'application/json' is explicitly passed.

bblimke:
"If a server is actually returning JSON as the body content" - which server do you mean? We are returning a stubbed response. This response will be passed through specific http client library internals. You can't expect these libs to handle Hash.

WebMock would have to convert Hash declared in `to_return` to json string automatically first, but how would it know whether to convert this Hash to json or to xml?

joshuaswilcox:
Would it not be best to use the headers content-type to determine whether to return JSON or XML? I am stubbing as so

    stub_request(:post,'http://......').
         to_return(
          :status => 200,
          :body => '{"transaction_status":"declined"}',
          :headers => {"Content-Type"=> "application/json"})

but would prefer not to have to call JSON.parse(response_body) since that is not a step that is required in my code as JSON is implicitly declared.  Maybe I am missing something though?  FWIW, I am using Faraday to make the actual request.

bblimke:
In case there is response content type declared it could work. Webmock would have to raise an error is case body is declared as a hash and headers are not declared.

bblimke:
Most people will forget about headers.

joshuaswilcox:
I don't think that is a problem to require a content-type when passing certain types of body content, since that is generally a good practice across http requests.  Could i make a pull request for this functionality?

bblimke:
[...]

PR #727, opened 2017-10-27
Allows JSON to be passed if JSON Content-Type

LDonoughe:
Doesn't support hashes in general

Re: https://github.com/bblimke/webmock/issues/449

### DIFF
--- a/lib/webmock/http_lib_adapters/typhoeus_hydra_adapter.rb
+++ b/lib/webmock/http_lib_adapters/typhoeus_hydra_adapter.rb
@@ -82,8 +82,8 @@ if defined?(Typhoeus)
         def self.build_webmock_response(typhoeus_response)
           webmock_response = WebMock::Response.new
           webmock_response.status = [typhoeus_response.code, typhoeus_response.status_message]
-          webmock_response.body = typhoeus_response.body
           webmock_response.headers = typhoeus_response.headers
+          webmock_response.body = typhoeus_response.body
           webmock_response
         end
 
--- a/README.md
+++ b/README.md
@@ -291,6 +291,16 @@ stub_request(:any, "www.example.com").
 Net::HTTP.get("www.example.com", '/')    # ===> "abc"
 ```
 
+### Stubbing with custom JSON response
+
+```ruby
+stub_request(:any, "www.example.com").
+  to_return(body: {abc: 'def'}, status: 200,
+    headers: { 'Content-Type' => 'application/json' })
+
+Net::HTTP.get("www.example.com", '/')    # ===> "{abc: 'def'}"
+```
+
 ### Response with body specified as IO object
 
 ```ruby
--- a/lib/webmock/http_lib_adapters/em_http_request_adapter.rb
+++ b/lib/webmock/http_lib_adapters/em_http_request_adapter.rb
@@ -201,7 +201,7 @@ if defined?(EventMachine::HttpClient)
         response_string = []
         response_string << "HTTP/1.1 #{status[0]} #{status[1]}"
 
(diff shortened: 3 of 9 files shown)


bblimke:
@LDonoughe thank you for the pull request!

This would indeed be a very useful addition to Webmock.

In order to support hashes as stubbed response body, Webmock would either need
to convert them to json or make sure all http client libs support hash as a response body.
I believe this has been raised in https://github.com/bblimke/webmock/issues/449

To make the pull request is complete, an appropriate acceptance spec would be useful,
to make sure that this functionality actually works across all http client libs,
and updated README.

If we allow WebMock to accept hashes just for JSON, then perhaps
Webmock should provide a more specific error message in case Hash is provided,
but json content type is not?

LDonoughe:
You're Welcome @bblimke !

I've tried to address your PR comments in my latest commit. The reason so many files were touched is because a number of the adaptors were relying on implicit string conversion where Hash does not have one, so I'm explicitly `to_s` on the body where I got errors. Let me know if these changes do not completely address your concerns and what I can do to change that.

lib/webmock/http_lib_adapters/typhoeus_hydra_adapter.rb
```diff
@@ -82,8 +82,8 @@ def self.build_request_signature(req)
         def self.build_webmock_response(typhoeus_response)
           webmock_response = WebMock::Response.new
           webmock_response.status = [typhoeus_response.code, typhoeus_response.status_message]
-          webmock_response.body = typhoeus_response.body
           webmock_response.headers = typhoeus_response.headers
+          webmock_response.body = typhoeus_response.body
```
LDonoughe: This was necessary due to the body validation relying on headers for Hash now

LDonoughe:
@bblimke is there anything else I can do to speed up the acceptance/merge of this PR? My team could really use this functionality to properly test 3rd-party endpoints

bblimke:
@LDonoughe sorry. I missed your latest changes.

Putting duplication aside, I'm not sure I understand the need for `to_s` change in all adapters.

Are you trying to convert the Hash object to String in order for http clients to accept it?
Hash `to_s` is not JSON.

Please see my above comment:

>In order to support hashes as stubbed response body, Webmock would either need
to convert them to json or make sure all http client libs support hash as a response body.
I believe this has been raised in #449 

The current specs also don't actually check the body returned from request.

LDonoughe:
@bblimke I know it's been a couple years but I wanted to share my findings and close this PR.

The short answer is that I was wrong about needing to return a hash rather than a string. So, my advice to anyone who finds this PR is to check what data is actually being returned by your endpoint via your choice of HTTP adapter. If it's actually a hash and not a string, feel free to resurrect my "fix." In most cases, you're going to have to JSON.parse the string to get your JSON.

As far as the integration tests go: when I attempted to add integration tests for the JSON work I noticed that it was broken in 4 or so HTTP adapters. When I went to add a similar shared test for Array to see why that was the case I noticed that that too was broken. I lost the original code/tests I had but I've created a new Array integration test if you'd like to see for yourself by checking out my branch (https://github.com/LDonoughe/webmock/tree/data_type_support_tests) and running `bundle exec rake spec`

I sincerely hope this saves at least one person some time and effort testing their integrations.

bblimke:
@LDonoughe thank you for these insights and the for closing the issue. Hopefully that will help others.

Yes, there are many trade-offs in WebMock in order to support various http clients,
but that's the price to have a unified API that works everywhere (in theory ;)

People can always mock the specific http clients in case WebMock is too generic.

PR #731, opened 2017-11-30
Automatically disable WebMock after rspec suite

d4rky-pl:
When using [ruby-codacy-coverage](https://github.com/codacy/ruby-codacy-coverage) we were forced to whitelist Codacy by hand, even though the code coverage was being sent long after the entire test suite has run (in `at_exit` block in [simplecov](https://github.com/colszowka/simplecov)).

This was a surprising behaviour as the point of requiring `webmock/rspec` is to have a seamless integration that doesn't interfere with other tooling.

This PR changes the default behaviour of `webmock/rspec` so it does not by default interfere with test configuration.

### DIFF
--- a/lib/webmock/rspec.rb
+++ b/lib/webmock/rspec.rb
@@ -20,12 +20,18 @@ end
 
 require 'webmock/rspec/matchers'
 
-WebMock.enable!
-
 RSPEC_CONFIGURER.configure { |config|
 
   config.include WebMock::API
   config.include WebMock::Matchers
+
+  config.before(:suite) do
+    WebMock.enable!
+  end
+
+  config.after(:suite) do
+    WebMock.disable!
+  end
 
   config.after(:each) do
     WebMock.reset!


bblimke:
@d4rky-pl thanks for the research and the pull request. That looks useful!

I'm trying to think if this change is sufficiently backwards compatible and if it can break an existing setups for anyone.

I believe this could only break if someone does the following currently:

````
require 'webmock/rspec'
WebMock.disable!
````

The other option is that someone takes advantage of WebMock being enabled,
before the `before(:suite)` callbacks are invoked.

Both scenarios are fairly unlikely, don't you think?

d4rky-pl:
I was hoping to start a discussion around this. Right now the current setup also makes sure that no network calls are called for the setup stage, duration of entire suite, _end then some_. For some setups it definitely makes sense to avoid any possible network calls (testing in total isolation, CI with no network adapter, making sure your tests are impervious to network issues etc). On the other hand, `require "webmock/rspec"` can make a person assume that this is only the rspec integration and not "disable everything for as long as process is running".

I was frankly more worried about the exact opposite case, which would be not being able to enable webmock for both test suite and additional tooling (for ex. for the previously mentioned testing in total isolation) - even turning webmock manually on will still disable it after running a full suite so to enable it you'd have to require just webmock and include webmock helpers manualy (or add some extra way of disabling the `after` callback / reenabling webmock)

I don't think there's a good or bad solution in here but my PR do change the default behaviour. It might be a good idea to either merge it on full minor/major version, add extra step to make it work the way I'm proposing (require + `Webmock.enable_rspec!`? `disable_after_suite`?) or at the very least add a disclaimer in README that says the rspec integration only works for the duration of test suite and not for setup files and after suite callbacks.

Choices, choices...

bblimke:
@d4rky-pl the chance of someone being affected by that change seems so low to me that I'm tempted to do a minor version release, since it's an improvement over how webmock/rspec worked so far, and is also a bug fix in a way.

d4rky-pl:
Yeah, after thinking this through I agree

PR #734, opened 2017-12-28
Add JSON API support

ouranos:
Add support for [JSON API](http://jsonapi.org/).

This makes `hash_including` works properly with JSON API requests.

Not sure if I need to change [this part too](https://github.com/bblimke/webmock/blob/135a44446ee6d323e9970f39424ab5c6ec3efcc0/lib/webmock/request_signature.rb#L41)?

I wasn't sure which specs to update, let me know if you need me to add anything else.


bblimke:
@ouranos thank you!

If we add support for [this](https://github.com/bblimke/webmock/blob/135a44446ee6d323e9970f39424ab5c6ec3efcc0/lib/webmock/request_signature.rb#L41) then the failure output on non matching requests will be nicer, therefore I'd be keen to add support for that.

ouranos:
Ok, I'll add it then.

Any ideas why the specs are failing? It doesn't seem to be related to my changes.

bblimke:
@ouranos please merge from master. I have fixed specs under jruby.

ouranos:
Rebased from `master` and added `RequestSignature` support. Let's see if the build pass.

hoenic07:
This PR should fix the JSON API support, in a more generic way (and allows further content type customization): https://github.com/bblimke/webmock/pull/763
Since JSON API is not that widely adopted yet, I wouldn't add it to the default content types.

PR #736, opened 2017-12-29
Typhoeus returns OK for a return code

spizm:
Whenever Typhoeus successfully completes a request, the return code is `:ok`. This pull request adds this return code to the adapter.

### DIFF
--- a/lib/webmock/http_lib_adapters/typhoeus_hydra_adapter.rb
+++ b/lib/webmock/http_lib_adapters/typhoeus_hydra_adapter.rb
@@ -102,6 +102,7 @@ if defined?(Typhoeus)
               status_message: webmock_response.status[1],
               body: webmock_response.body,
               headers: webmock_response.headers,
+              return_code: :ok,
               effective_url: request_signature.uri
             )
           end


bblimke:
@spizm thank you, but I believe this pull request is not complete.

According to Typhoeus code, :ok is only returned for 200-299 responses: https://github.com/typhoeus/typhoeus/blob/d03fceba09d0296b94ed009fb1b5c0cfc95e7418/spec/typhoeus/response/status_spec.rb#L78

If we are to add support for return_code, then I guess we should handle all other cases as well.

nrw505:
Typhoeus sets `return_code: :ok` for HTTP statuses other than 200-299

e.g.

```
irb(main):001> require 'typhoeus'
=> true
irb(main):002> response = Typhoeus.get("http://localhost:5005/500")
=> 
#<Typhoeus::Response:0x00007f862c8e7b78
...
irb(main):003> response.return_code
=> :ok
irb(main):004> response.code
=> 500
irb(main):005> 
```

bblimke:
@nrw505 thank you for reporting. Do you know what are the other return_code's and for what status codes should these be returned?

PR #738, opened 2018-01-03
Remove unused `pos` variable to fix Ruby warning

mattbrictson:
This fixes the following warning that is printed when using Ruby 2.5.0:

```
lib/webmock/util/json.rb:27: warning: assigned but unused variable - pos
```


bblimke:
Thank you!

bblimke:
Released as 3.2.1

Issue #489, opened 2015-06-18
Invalid scheme format. (Addressable::URI::InvalidURIError)


beanieboi:
Hi,

i get the Invalid scheme format. (Addressable::URI::InvalidURIError) error when i use templates with Webmock. i traced the error down this this line https://github.com/bblimke/webmock/blob/master/lib/webmock/request_pattern.rb#L136

`uri = "http://127.0.0.1:50153/__identify__"`
and where 
`WebMock::Util::URI.variations_of_uri_as_strings(uri)`
creates the following array:
`["http://127.0.0.1:50160/__identify__", "127.0.0.1:50160/__identify__"]`
passing that to `any?` where the `@pattern.match(u)` throws the exception when comparing with the second element from the hash.

i looked up `WebMock::Util::URI.variations_of_uri_as_strings(uri)` and it creates them on purpose.

```
if normalized_uri.scheme == "http"
  uris = uris_with_scheme_and_without(uris)
end
```

so i assume that array of URI's is correct. what other thing could go wrong here?
can you point me at something to debug?

_UPDATE_

thats not the URI i want to template.

```
bitbucket_uri = Addressable::Template.new("https://bitbucket.org/api/1.0/repositories/codeship/{bitbucket_repo_name}/services")
  WebMock.stub_request(:get, bitbucket_uri).
    to_return(status: 200)
```

i guess thats coming from some other tool in our cucumber suite. but if i remove any template from Webmock, the error is gone.
thanks
ben

ngauthier:
I believe this is an issue with Addressable's parser:

```
[1] pry(main)> Addressable::URI.parse("http://127.0.0.1/")
=> #<Addressable::URI:0x3fc2960f6df4 URI:http://127.0.0.1/>
[2] pry(main)> Addressable::URI.parse("127.0.0.1/")
=> #<Addressable::URI:0x3fc2960c93f4 URI:127.0.0.1/>
[3] pry(main)> Addressable::URI.parse("127.0.0.1/_")
=> #<Addressable::URI:0x3fc296098ce0 URI:127.0.0.1/_>
[4] pry(main)> Addressable::URI.parse("127.0.0.1/__")
=> #<Addressable::URI:0x3fc29606f8a4 URI:127.0.0.1/__>
[5] pry(main)> Addressable::URI.parse("127.0.0.1:123")
Addressable::URI::InvalidURIError: Invalid scheme format.
[6] pry(main)> Addressable::URI.parse("foo:123")
[...]

PR #739, opened 2018-01-04
Fix for "Invalid scheme format"

dwbutler:
Fixes https://github.com/bblimke/webmock/issues/489. See https://github.com/sporkmonger/addressable/issues/208

### DIFF
--- a/lib/webmock/request_pattern.rb
+++ b/lib/webmock/request_pattern.rb
@@ -155,10 +155,10 @@ module WebMock
     def matches?(uri)
       if @query_params.nil?
         # Let Addressable check the whole URI
-        WebMock::Util::URI.variations_of_uri_as_strings(uri).any? { |u| @pattern.match(u) }
+        WebMock::Util::URI.variations_of_uri_as_strings(uri).any? { |u| @pattern.match(WebMock::Util::URI.heuristic_parse(u)) }
       else
         # WebMock checks the query, Addressable checks everything else
-        WebMock::Util::URI.variations_of_uri_as_strings(uri.omit(:query)).any? { |u| @pattern.match(u) } &&
+        WebMock::Util::URI.variations_of_uri_as_strings(uri.omit(:query)).any? { |u| @pattern.match(WebMock::Util::URI.heuristic_parse(u)) } &&
           @query_params == WebMock::Util::QueryMapper.query_to_values(uri.query)
       end
     end


bblimke:
Thanks.

Most likely the result of heuristic parse will be the same for all variations of the url string, therefore multiple checks might be redundant. I need to think about it.
On the other side I'm not sure if precision won't be lost when converting the string to the url created with heuristic_parse. 
All specs pass, but I need to make sure that none are missing.

skunkworker:
:+1:  on the patch, it fixed the issues I was having with an addressable template and webmock.

tmertens:
@bblimke any update on this?

oliakremmyda:
The following specs are failing:
```
rspec ./spec/unit/request_pattern_spec.rb:114 # WebMock::RequestPattern when matching should match if uri Addressable::Template pattern matches unescaped form of request uri
rspec ./spec/unit/request_pattern_spec.rb:119 # WebMock::RequestPattern when matching should match if uri Addressable::Template pattern matches request uri
rspec ./spec/unit/request_pattern_spec.rb:161 # WebMock::RequestPattern when matching when parameters are escaped should match if uri Addressable::Template pattern matches uri without parameter value and request has escaped parameters
rspec ./spec/unit/request_pattern_spec.rb:166 # WebMock::RequestPattern when matching when parameters are escaped should match if uri Addressable::Template pattern matches uri without parameter value and request has unescaped parameters
rspec ./spec/unit/request_pattern_spec.rb:171 # WebMock::RequestPattern when matching when parameters are escaped should match if uri Addressable::Template pattern matches uri with unescaped parameter value and request has unescaped parameters
rspec ./spec/unit/request_pattern_spec.rb:176 # WebMock::RequestPattern when matching when parameters are escaped should match if uri Addressable::Template pattern matches uri with escaped parameter value and request has escaped parameters
rspec ./spec/unit/request_pattern_spec.rb:222 # WebMock::RequestPattern when matching when matching requests on query params when uri is described as Addressable::Template should raise error if query params are specified
rspec ./spec/unit/request_pattern_spec.rb:232 # WebMock::RequestPattern when matching when matching requests on query params when uri is described as Addressable::Template should match when query params are declared as HashIncluding matcher matching params
rspec ./spec/unit/request_pattern_spec.rb:244 # WebMock::RequestPattern when matching when matching requests on query params when uri is described as Addressable::Template should match when query params are declared as RSpec HashIncluding matcher matching params
```

PR #754, opened 2018-04-05
Fix redundant `Util::` namespacing

EiNSTeiN-:
On line 5 the reference to `Util` resolves to the class that was just opened on line 4, so `Util::` is redundant.

FYI @bblimke


bblimke:
Thanks @EiNSTeiN-  👍

PR #758, opened 2018-05-19
Fix invalid scheme error with Addressable::Template

### LINKED ISSUE #489: Invalid scheme format. (Addressable::URI::InvalidURIError)

uiureo:
Fix #489 

# problem
`Addressable::Template#match` expects a valid uri with scheme
  https://github.com/sporkmonger/addressable/issues/208#issuecomment-162404607

It works with most cases without uri scheme, but it raises an error with some cases like `127.0.0.1:8080`.

I've found that it causes the error when using webmock with [System Spec](https://relishapp.com/rspec/rspec-rails/docs/system-specs/system-spec) and chrome driver.

# solution
Always pass a uri with scheme to `Addressable::Template#match`. 

The difference from #739 is 
- adding a spec 
- making failing specs pass

### DIFF
--- a/lib/webmock/util/uri.rb
+++ b/lib/webmock/util/uri.rb
@@ -33,7 +33,7 @@ module WebMock
         NORMALIZED_URIS[uri].dup
       end
 
-      def self.variations_of_uri_as_strings(uri_object, with_scheme: false)
+      def self.variations_of_uri_as_strings(uri_object, only_with_scheme: false)
         normalized_uri = normalize_uri(uri_object.dup).freeze
         uris = [ normalized_uri ]
 
@@ -47,7 +47,7 @@ module WebMock
           uris = uris_with_inferred_port_and_without(uris)
         end
 
-        if normalized_uri.scheme == "http" && !with_scheme
+        if normalized_uri.scheme == "http" && !only_with_scheme
           uris = uris_with_scheme_and_without(uris)
         end
 
--- a/lib/webmock/util/uri.rb
+++ b/lib/webmock/util/uri.rb
@@ -33,7 +33,7 @@ module WebMock
         NORMALIZED_URIS[uri].dup
       end
 
-      def self.variations_of_uri_as_strings(uri_object)
+      def self.variations_of_uri_as_strings(uri_object, with_scheme: false)
         normalized_uri = normalize_uri(uri_object.dup).freeze
         uris = [ normalized_uri ]
 
@@ -47,7 +47,7 @@ module WebMock
           uris = uris_with_inferred_port_and_without(uris)
         end
 
-        if normalized_uri.scheme == "http"
+        if normalized_uri.scheme == "http" && !with_scheme
           uris = uris_with_scheme_and_without(uris)
         end
 
--- a/lib/webmock/request_pattern.rb
+++ b/lib/webmock/request_pattern.rb
@@ -155,10 +155,10 @@ module WebMock
     def matches?(uri)
       if @query_params.nil?
         # Let Addressable check the whole URI
-        WebMock::Util::URI.variations_of_uri_as_strings(uri).any? { |u| @pattern.match(u) }
+        matches_with_variations?(uri)
       else
         # WebMock checks the query, Addressable checks everything else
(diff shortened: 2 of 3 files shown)


bblimke:
@uiureo thank you for the PR.

Could you please explain why have you removed all these urls in https://github.com/bblimke/webmock/pull/758/files#diff-a5f6621858e36a69278262b16c576604 ?

These urls were there to guarantee that WebMock can match urls without scheme provided with urls with the default scheme, which is not the case anymore.
As I understand this change breaks backwards compatibility?

uiureo:
Ah I understand. This is mistake.
I will fix it in a way that it doesn't change `WebMock::Util::URI.variations_of_uri_as_strings`.

uiureo:
Added `with_scheme: true` option to `WebMock::Util::URI.variations_of_uri_as_strings` instead.

@bblimke How do you think?

lib/webmock/util/uri.rb
```diff
@@ -47,7 +47,7 @@ def self.variations_of_uri_as_strings(uri_object)
           uris = uris_with_inferred_port_and_without(uris)
         end
 
-        if normalized_uri.scheme == "http"
+        if normalized_uri.scheme == "http" && !with_scheme
```
bblimke: I'm not sure I understand the with_scheme flag here. The condition is "if not with scheme" and the line below compares returns values with scheme and without.

uiureo: The flag means "returns only variations with scheme", but I agree it's confusing.

`only_variations_with_scheme` might be better?

bblimke: Yes `only_with_scheme` would be more meaningful.

uiureo: updated https://github.com/bblimke/webmock/pull/758/commits/19624000ba9701ade80df97a5d92d3bc3c2ca4a6

uiureo:
@bblimke Could you review this again?

bblimke:
@uiureo sorry. Looks good. Thank you!

bblimke:
Released as 3.4.2. Thank you!

alain-andre:
Hi there, I'm having this issue and I'm wondering why I cannot see this fix in the latest code : https://github.com/bblimke/webmock/blob/904aa9806a91fb292a142dc2783ff762eacca0d8/lib/webmock/request_pattern.rb#L190
It has been remove on [commit](https://github.com/bblimke/webmock/commit/8b1414850ed5b031bba0c21b1033058c4725ff2f)

I do not understand why in the UT it adds the `only_with_scheme` of the `WebMock::Util::URI.variations_of_uri_as_strings` method but removes it from the `request_pattern` class.

Any idea ?

bblimke:
@alain-andre it has been removed due to this issue https://github.com/bblimke/webmock/issues/828 

The problem was that the only_with_scheme solution is not complete.

Hopefully the issue will be resolved once https://github.com/bblimke/webmock/pull/890 is finished. @guppy0356

PR #751, opened 2018-03-15
Prevent `Net/ReadTimeout` error in Ruby 2.6

koic:
This PR fixes the following build error in Ruby 2.6.0 (ruby-head).

```console
$ ruby --version
ruby 2.6.0dev (2018-03-07 trunk 62693) [x86_64-linux]
% bundle exec rake
(snip)

  1) Net:HTTP with WebMock when net connect is allowed should make a
  real https request if request is not stubbed
     Failure/Error: response = super(request, nil, &nil)

     Net::ReadTimeout:
       Net::ReadTimeout
     Shared Example Group: "allowing and disabling net connect" called
       from ./spec/acceptance/webmock_shared.rb:25
     Shared Example Group: "with WebMock" called from
       ./spec/acceptance/net_http/net_http_spec.rb:10
     # ./lib/webmock/http_lib_adapters/net_http.rb:97:in `block in
       request'
     # ./lib/webmock/http_lib_adapters/net_http.rb:105:in `block in
       request'
     # ./lib/webmock/http_lib_adapters/net_http.rb:137:in
       `start_with_connect_without_finish'
     # ./lib/webmock/http_lib_adapters/net_http.rb:104:in `request'
     # ./spec/acceptance/net_http/net_http_spec_helper.rb:35:in `block
       in http_request'
     # ./lib/webmock/http_lib_adapters/net_http.rb:123:in
       `start_without_connect'
     # ./lib/webmock/http_lib_adapters/net_http.rb:150:in `start'
     # ./spec/acceptance/net_http/net_http_spec_helper.rb:34:in
       `http_request'
     #
       ./spec/acceptance/shared/allowing_and_disabling_net_connect.rb:14:in
       `block (4 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # Net::ReadTimeout:
     #   Net::ReadTimeout
     #   ./lib/webmock/http_lib_adapters/net_http.rb:97:in `block in
       request'

(Errors continue after this)
```

https://travis-ci.org/bblimke/webmock/jobs/350668167#L942-L1202

This is because it was changed to use `IO#read_nonblock` based on the value of `@rbuf` by the following commit.
https://github.com/ruby/ruby/commit/b02fc0f9fedd635eb64f4bb6440ab63923c2af24

This PR forces the 2nd argument `tmp` of `IO#read_nonblock` to `nil` by using `TracePoint`. So this prevents `Net::ReadTimeout` from occurring.

PR #759, opened 2018-05-20
Only apply TracePoint to current thread

arkadiyt:
Hi,

While upgrading to webmock 3.4.0 I found that some of my specs started failing. The way my tests are setup, in addition to using webmock for stubbing out requests in unit tests I also perform integration tests by making unmocked http requests against a WEBRick server running in a different thread in my rspec process.

After a bit of debugging I've narrowed the issue down to the TracePoint code added in https://github.com/bblimke/webmock/pull/751 not being thread safe. In particular that code will undefine any local variable called `tmp` whenever any line of code is run in any thread, even not in the `rbuf_fill` thread for which it was intended.

The end result is I would get exceptions in places like [webrick/log.rb#L152](https://github.com/ruby/ruby/blob/trunk/lib/webrick/log.rb#L152):
```ruby
    def log(level, data)
      tmp = Time.now.strftime(@time_format)
      tmp << " " << data  # Throws exception here
      super(level, tmp)
    end
```
With errors on `tmp << " "` like `undefined method '<<' for nil:NilClass (NoMethodError)` (despite `tmp` having been immediately defined the line before).

Or [webrick/httputils.rb#L211](https://github.com/ruby/ruby/blob/trunk/lib/webrick/httputils.rb#L211):
```ruby
    def parse_qvalues(value)
      tmp = []
      if value
        parts = value.split(/,\s*/)
        parts.each {|part|
          if m = %r{^([^\s,]+?)(?:;\s*q=(\d+(?:\.\d+)?))?$}.match(part)
            val = m[1]
            q = (m[2] or 1).to_f
            tmp.push([val, q])  # Throws nil exception for undefined tmp here
```

My fix is to apply the tracepoint only to the `rbuf_fill` thread - after applying this change locally all my tests pass without exception.

Thanks

**edit:** I believe the `jruby-9.0.5.0` failure on travis is unrelated to my change

### DIFF
--- a/lib/webmock/http_lib_adapters/net_http.rb
+++ b/lib/webmock/http_lib_adapters/net_http.rb
@@ -274,7 +274,10 @@ module Net  #:nodoc: all
 
     if RUBY_VERSION >= '2.6.0'
       def rbuf_fill
+        current_thread_id = Thread.current.object_id
+
         trace = TracePoint.trace(:line) do |tp|
+          next unless Thread.current.object_id == current_thread_id
           if tp.binding.local_variable_defined?(:tmp)
             tp.binding.local_variable_set(:tmp, nil)
           end


arkadiyt:
@bblimke thanks for merging - would it be possible to cut a new patch release for this?

bblimke:
@arkadiyt Done. Thank you!

PR #761, opened 2018-06-04
Add API to reset registry requests

### LINKED ISSUE #760: Reset only history (not stubs)
I have run into a few times where I want to only reset the `execute count` and not "un-load" the stubs.

For example: a model where API calls are made on create and update. 
To test the update calls, I first create test records (e.g. FactoryBot). But since creation also triggers the API, my execute count is inflated before my test (in this case both create and update rely on POST, so I can't filter them out).

Using this `WebMock::RequestRegistry.instance.reset!` works (with maybe no unintended consequences?), but it might be nice to have a `WebMock.reset_history!` or `WebMock.reset_count!` so it feels a bit more intentionally available.

I'd offer a PR, but I'm a hobbyist and only understand about half of what I see.

oliakremmyda:
Fixes #760

### DIFF
--- a/README.md
+++ b/README.md
@@ -460,6 +460,37 @@ stub_get = stub_request(:get, "www.example.com")
 remove_request_stub(stub_get)
 ```
 
+### Reset global request counter
+
+```ruby
+stub_get = stub_request(:get, "www.example.com")
+Net::HTTP.get('www.example.com', '/')
+
+WebMock::RequestRegistry.instance.times_executed(stub_get.request_pattern)  # => 1
+reset_executed_requests!
+WebMock::RequestRegistry.instance.times_executed(stub_get.request_pattern)  # => 0
+```
+
+but also:
+
+```ruby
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -460,6 +460,37 @@ stub_get = stub_request(:get, "www.example.com")
 remove_request_stub(stub_get)
 ```
 
+### Reset global request counter
+
+```ruby
+stub_get = stub_request(:get, "www.example.com")
+Net::HTTP.get('www.example.com', '/')
+
+WebMock::RequestRegistry.instance.times_executed(stub_get.request_pattern)  # => 1
```
bblimke: `WebMock::RequestRegistry.instance.times_executed` is not part of the API.

perhaps it's better to show examples with `expect(stub_get).to have_been_requested.times(1)` 
and then `expect(stub_get).not_to have_been_requested` ?

oliakremmyda: Sure, I will change this!

PR #763, opened 2018-06-07
ADD configurable custom content type mapping for request body matching

hoenic07:
jsonapi (http://jsonapi.org/format/) requests have the custom content type `application/vnd.api+json`, which is not recognized by the current `BODY_FORMATS` as a `json` request. Since others might have similar issues with this, adding `application/vnd.api+json` to the `BODY_FORMATS` would have only solved the problem by it's root. Therefore I've added the possibility to customize the possible content types and to which format they map.

### DIFF
--- a/lib/webmock/webmock.rb
+++ b/lib/webmock/webmock.rb
@@ -83,7 +83,7 @@ module WebMock
     end
   end
 
-  def self.custom_content_type_mapping(config)
+  def self.custom_content_type_mapping=(config)
     Config.instance.custom_content_type_mapping = config
   end
 
--- a/lib/webmock/webmock.rb
+++ b/lib/webmock/webmock.rb
@@ -83,6 +83,10 @@ module WebMock
     end
   end
 
+  def self.custom_content_type_mapping(config)
+    Config.instance.custom_content_type_mapping = config
+  end
+
   def self.show_body_diff!
     Config.instance.show_body_diff = true
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+###
+
+* Add ability to config custom content type mapping for request body matching
+
 ## 3.4.2
 
   * Fixed `rbuf_fill` in Net::HTTP adapter to be thread-safe
--- a/README.md
+++ b/README.md
@@ -158,10 +158,10 @@ RestClient.post('www.example.com', '<data a="1" b="five" />',
   content_type: 'application/xml')    # ===> Success
 ```
 
-With a custom content type mapping:
+With an additional custom content type mapping:
 
 ```ruby
-WebMock.custom_content_type_mapping("application/vnd.api+json" => :json)
+WebMock.custom_content_type_mapping = { "application/vnd.api+json" => :json }
 
 stub_request(:post, "www.example.com").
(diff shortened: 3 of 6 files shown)


bblimke:
@hoenic07 thank you for this PR.

Is there a good not to map `application/vnd.api+json` to :json by default? If it doesn't make sense to map `application/vnd.api+json` to anthing else than json, then perhaps it's not worth bothering WebMock users with having to add the configuration?

hoenic07:
@bblimke I also thought of this. But there are other json content types (e.g. `application/geo+json` or https://www.iana.org/assignments/media-types/media-types.xhtml) which people might need as well. We could of course add all of them in separate PRs, but my PR adds the flexibility to be able to include the ones needed. Most of the time the default `application/json` is fine, but sometimes people need a custom  sub type.

bblimke:
@hoenic07 The configuration you have added is valuable, but despite that, do you think `application/vnd.api+json` is popular enough to be included by default?

hoenic07:
@bblimke Since there are no other non-standard types in `BODY_FORMATS` and jsonapi isn't going to be super popular in the near future I wouldn't add it to the default to be honest.

lib/webmock/webmock.rb
```diff
@@ -83,6 +83,10 @@ def self.net_connect_explicit_allowed?(allowed, uri=nil)
     end
   end
 
+  def self.custom_content_type_mapping(config)
```
bblimke: shouldn't that method be named `custom_content_type_mapping=` if it's a setter?

seisen77:
I could really use this right about now.

johannesluedke:
This would be really useful for all the projects that use JSON-API 😎 (and others). It looks like it's working - Is there anything that prevents this from being merged & released that one could help with :) ? @bblimke

bblimke:
@johannesluedke what is the content type that you use?

bblimke:
@seisen77 sorry that was 2 years ago, but what is the content type that you used?

I'm wondering how how likely it is that people will need to add custom content types if Webmock included a few most popular json content types by default, instead of requiring users to add configuration.

johannesluedke:
> what is the content type that you use?

We're using `'application/vnd.api+json'` - that's also the one used in the official jsonapi documentation at https://jsonapi.org/format/ 👍

bblimke:
looking at https://www.iana.org/assignments/media-types/media-types.xhtml it seems to me that all json custom types have this pattern `\Aapplication\/[a-zA-Z0-9.-]+\+json\Z` (I know this regex is not strictly accurate and covers more than it should, but just for simplicity :)
Perhaps instead of requiring users to configure things, Webmock could be smart and recognise json by this pattern.
What do you think?

johannesluedke:
@bblimke I would be fine with that :)

bblimke:
Ok, the solution that doesn't require any custom configuration and works for all \Aapplication\/[a-zA-Z0-9.-]+\+(json|xml)\Z content types is now released as version 3.12.

@johannesluedke thank you for your feedback.

@hoenic07 thank you for your PR. I may reopen it if there is a need in the future to have custom content types,
that don't follow conventional naming that could be solved with a regexp pattern.

johannesluedke:
Thanks for the fast release 🙏 😇

PR #766, opened 2018-06-18
Bump addressable version

mogox:
Addressable 2.5+ adds support for ruby 2.4 which also is a dependency in other gems.


bblimke:
@mogox how does previous constraint `'>= 2.3.6'` prevent your form using Addressable 2.5?

mogox:
It didn't! Sorry about that, it was another dependency that was stopping my bundle.  Thanks for the quick response

PR #771, opened 2018-08-21
Travis: use latest patch releases

olleolleolle:
This PR only updates the versions of Rubies in CI.

See https://github.com/rbenv/ruby-build/tree/master/share/ruby-build for versions used.


bblimke:
@olleolleolle Thank you! :)

PR #776, opened 2018-09-05
replace block.call with yield

ahorek:


bblimke:
@ahorek could you please explain why you believe yield is better here?

ahorek:
hi @bblimke
here's a good explanation
https://github.com/JuanitoFatas/fast-ruby#proccall-and-block-arguments-vs-yieldcode
and a benchmark
https://github.com/JuanitoFatas/fast-ruby/blob/master/code/proc-and-block/proc-call-vs-yield.rb

this was improved in ruby 2.5 and upcomming ruby 2.6, but yield is still much faster (about 5 times) then block.call
https://bugs.ruby-lang.org/issues/14329

PR #780, opened 2018-09-14
Fix encoding issue when using assert_requested during test runs where images have been uploaded.

adamgaynor:


bblimke:
@adamgaynor thank you. The change looks useful, though I'm not sure if modifying request_signature objects is a good idea, just because printing the signature failed.

adamgaynor:
@bblimke It looked like modifying the `request_signature` was the simplest way to allow webmock to function again and properly analyze requests, when writing tests.
Even if there were a way to properly print the signature, you'd end up with an enormous wall of encoded image (in this situation).
I've been using a branch with this change for my own project.

oliakremmyda:
Sorry but the issue is not so clear to me, @adamgaynor can you please add an example on the description or/and a couple of specs to document it? Thank you

PR #788, opened 2018-12-12
disable TracePoint properly

ko1:
if `super` raise an exception, this very danger TracePoint is not disabled.
This is why https://bugs.ruby-lang.org/issues/15400

This change is introduced at: https://github.com/bblimke/webmock/pull/751

Anyway, this approach is very danger (for example, `tmp` variable name is not exposed API), so that pls re-consider to use this technique.

### DIFF
--- a/lib/webmock/http_lib_adapters/net_http.rb
+++ b/lib/webmock/http_lib_adapters/net_http.rb
@@ -285,7 +285,7 @@ module Net  #:nodoc: all
         end
 
         super
-
+      ensure
         trace.disable
       end
     end


lib/webmock/http_lib_adapters/net_http.rb
```diff
@@ -285,7 +285,7 @@ def rbuf_fill
         end
 
         super
-
+      ensure
```
hanachin: In Ruby 2.6, the `TracePoint#enable` takes `target:` keyword argument.
We can use `TracePoint#enable` with the `target:` to enable the TracePoint for the specific method.
What do you think?
```rb
        trace.enable(target: WebMock::HttpLibAdapters::NetHttpAdapter::OriginalNetBufferedIO.instance_method(:rbuf_fill)) do
          super
        end
```

ko1: you == me?

It is better. But I think this TracePoint approach is danger, even with this limitation.

bblimke:
Any thoughts of how to avoid using trace point to achieve the same behaviour?

ko1:
Not sure because I don't know about `net/protocol.rb`, but modifying `@rbuf`? monkey patching is better than it because you need to check the behavior change (maybe you specify the version more explicitly).

lamont-granquist:
ping?  it might be better to remove tracepoint, but merging this would unblock our 2.6 testing.

Issue #774, opened 2018-08-30
Proposal to alias `enable_net_connect!` and `disallow_net_connect!`


owyongsk:
Currently there's

`WebMock.allow_net_connect!` and `WebMock.disable_net_connect!`. For readability and memorability we should add `enable_net_connect!` and `disallow_net_connect` because `enable` suggests it should be `disable`.

or stick with one of them at least. What do you guys think?

mykhi:
+1 from me, do not see why an opposite of "disable" is "allow".

bblimke:
@owyongsk makes sense. Do you fancy a pull request?

owyongsk:
Will take a crack at it. Thanks!

owyongsk:
The pull request is made, care to take a look? @bblimke

PR #789, opened 2018-12-17
Add aliases of enable_net_connect and disallow_net_connect

owyongsk:
https://github.com/bblimke/webmock/issues/774

Currently there's `WebMock.allow_net_connect!` and `WebMock.disable_net_connect!`. 

For readability and memorability we should add `enable_net_connect!` and `disallow_net_connect` because enable suggests it should be disable.

### DIFF
--- a/spec/unit/webmock_spec.rb
+++ b/spec/unit/webmock_spec.rb
@@ -10,10 +10,10 @@ describe "WebMock version" do
   end
 
   it "should alias enable_net_connect! to allow_net_connect!" do
-    expect(WebMock.enable_net_connect!).to eq(WebMock.allow_net_connect!)
+    expect(WebMock.method(:enable_net_connect!)).to eq(WebMock.method(:allow_net_connect!))
   end
 
   it "should alias disallow_net_connect! to disable_net_connect!" do
-    expect(WebMock.disallow_net_connect!).to eq(WebMock.disallow_net_connect!)
+    expect(WebMock.method(:disallow_net_connect!)).to eq(WebMock.method(:disallow_net_connect!))
   end
 end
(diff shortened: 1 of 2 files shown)


spec/unit/webmock_spec.rb
```diff
@@ -8,4 +8,12 @@
   it "should not require safe_yaml" do
     expect(defined?SafeYAML).to eq(nil)
   end
+
+  it "should alias enable_net_connect! to allow_net_connect!" do
+    expect(WebMock.enable_net_connect!).to eq(WebMock.allow_net_connect!)
```
bblimke: @owyongsk is that the right way to test that? is that not comparing the result of these two method invocations?

owyongsk: Makes sense, I've changed it to using `Object.method`.

spec/unit/webmock_spec.rb
```diff
@@ -10,10 +10,10 @@
   end
 
   it "should alias enable_net_connect! to allow_net_connect!" do
-    expect(WebMock.enable_net_connect!).to eq(WebMock.allow_net_connect!)
+    expect(WebMock.method(:enable_net_connect!)).to eq(WebMock.method(:allow_net_connect!))
   end
 
   it "should alias disallow_net_connect! to disable_net_connect!" do
-    expect(WebMock.disallow_net_connect!).to eq(WebMock.disallow_net_connect!)
+    expect(WebMock.method(:disallow_net_connect!)).to eq(WebMock.method(:disallow_net_connect!))
```
bblimke: @owyongsk comprating disallow to disallow?

owyongsk: Geez, I'm sorry. ):

bblimke:
Looks good. Thank you @owyongsk 👍

PR #796, opened 2018-12-29
[ready] perform should reset webmock_method after making request

TienDo1011:
I feel like each request should not care about what is the previously called HTTP verb.
For example:
```
c = Curl::Easy.new("http://www.google.co.uk")
c.http_post
c.perform
```
In this case, c.perform http_verb would be set automatically to POST, which is an surprise. If it's default to GET, it feels more intuitive. Since we're already familiar with the notion of stateless request, which is independent from each other.

### DIFF
--- a/lib/webmock/http_lib_adapters/curb_adapter.rb
+++ b/lib/webmock/http_lib_adapters/curb_adapter.rb
@@ -54,6 +54,7 @@ if defined?(Curl)
 
   module Curl
     class WebMockCurlEasy < Curl::Easy
+      attr_reader :webmock_method
 
       def curb_or_webmock
         request_signature = build_request_signature
--- a/lib/webmock/http_lib_adapters/curb_adapter.rb
+++ b/lib/webmock/http_lib_adapters/curb_adapter.rb
@@ -272,6 +272,7 @@ if defined?(Curl)
       def perform
         @webmock_method ||= :get
         curb_or_webmock { super }
+      ensure
         reset_webmock_method
       end
 
--- a/lib/webmock/http_lib_adapters/curb_adapter.rb
+++ b/lib/webmock/http_lib_adapters/curb_adapter.rb
@@ -271,6 +271,7 @@ if defined?(Curl)
       def perform
         @webmock_method ||= :get
         curb_or_webmock { super }
+        reset_webmock_method
       end
 
       def put_data= data
@@ -331,6 +332,10 @@ if defined?(Curl)
             super
           end
         METHOD
+      end
+
+      def reset_webmock_method
+        @webmock_method = :get
       end
 
       def reset
--- a/lib/webmock/http_lib_adapters/curb_adapter.rb
+++ b/lib/webmock/http_lib_adapters/curb_adapter.rb
@@ -54,8 +54,6 @@ if defined?(Curl)
 
   module Curl
     class WebMockCurlEasy < Curl::Easy
-      attr_reader :webmock_method
-
       def curb_or_webmock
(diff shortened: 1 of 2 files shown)


bblimke:
@TienDo1011 thank you.

As you can see your change has broken most Curb specs as `perform` is used internally by other methods.

There is a bug indeed that the previous invocation of `perform` leaves a dirty state on the object,
and this state needs to be reset before each http request.

Please let me know if you'd like to provide a complete PR with a spec example for curb adapter,
and the proper solution that works for all Curl::Easy methods.

TienDo1011:
@bblimke, Yeah I'd like to. So from what I understand, I'd need to:
- Find a solution to reset perform, so that it has a clean state prior to any new request.
- Fix failed spec test.

Sorry if this sounds a bit strange, since I'm new to contributing to open source :).

bblimke:
@TienDo1011 yes, and add a new example in curb_spec.rb to make sure the bug is fixed.

TienDo1011:
Got it @bblimke.

TienDo1011:
@bblimke I added a fix & an example to cover the fix. Can you take a look?
The failed spec on CI was coming from master.
Thank you very much.

lib/webmock/http_lib_adapters/curb_adapter.rb
```diff
@@ -333,6 +335,10 @@ def on_#{callback} &block
         METHOD
       end
 
+      def reset_webmock_method
+        @webmock_method = :get
```
bblimke: why :get and not nil?

lib/webmock/http_lib_adapters/curb_adapter.rb
```diff
@@ -271,6 +272,7 @@ def http_post *data
       def perform
         @webmock_method ||= :get
         curb_or_webmock { super }
+        reset_webmock_method
```
bblimke: This won't be called in case line above raises an error.

lib/webmock/http_lib_adapters/curb_adapter.rb
TienDo1011: I think putting user in a `Pit of success` is always better. Since we all assume default method would be `:get` if we not specify it otherwise anyway.

lib/webmock/http_lib_adapters/curb_adapter.rb
```diff
@@ -54,6 +54,7 @@ def self.parse_header_string(header_string)
 
   module Curl
     class WebMockCurlEasy < Curl::Easy
+      attr_reader :webmock_method
```
bblimke: is there a need for that reader?

spec/acceptance/curb/curb_spec.rb
```diff
@@ -411,6 +411,16 @@
...
       include CurbSpecHelper::NamedHttp
 
+      it "should reset @webmock_method after each call" do
+        stub_request(:post, "www.example.com").with(body: "01234")
+        c = Curl::Easy.new
+        c.url = "http://www.example.com"
+        c.post_body = "01234"
+        c.http_post
+        expect(c.response_code).to eq(200)
+        expect(c.webmock_method).to eq(:get)
```
bblimke: I don't think we should test internals. We can simply do perform and validate that it was a get request.

TienDo1011: Yeah you're right. I updated the code @bblimke.

bblimke:
Everything looks good. Thank you @TienDo1011

PR #799, opened 2019-01-02
Demonstrate PR 758 caused a regression

valscion:
DO NOT MERGE

This PR contains the new test added in https://github.com/bblimke/webmock/pull/797 followed by reverse of #758. This demonstrates that the fix done in #758 caused the error I talked about in #797.

Let's keep the discussion in #797 so all comments are in one place ☺️

### DIFF
--- a/lib/webmock/request_pattern.rb
+++ b/lib/webmock/request_pattern.rb
@@ -155,10 +155,10 @@ module WebMock
     def matches?(uri)
       if @query_params.nil?
         # Let Addressable check the whole URI
-        matches_with_variations?(uri)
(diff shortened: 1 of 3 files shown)


valscion:
Hmm seems like the Ruby 2.6.0 test matrix failure is also happening in master. Other specs do pass, though:

* Ruby 2.3.8 and this PR is green: https://travis-ci.org/bblimke/webmock/jobs/474417154
   
   screen shot 2019-01-02 at 16 11 06

* Ruby 2.3.8 and only the new spec is red: https://travis-ci.org/bblimke/webmock/jobs/474411068
  
  screen shot 2019-01-02 at 16 10 22

PR #804, opened 2019-01-28
Add support for net-http2 backend

singpolyma-shopify:


bblimke:
@singpolyma-shopify Thank you! Do you plan to add specs and cover real requests?

singpolyma-shopify:
@bblimke I'd like to.  I worked on this a bit just now, but I think the test server for testing "real" requests might not support http/2, which make re-using the existing acceptance tests a bit tricky.

bblimke:
@singpolyma-shopify right, that would probably require a new set of tests.

PR #809, opened 2019-03-05
Remove shebang from Rakefile

pvalena:
and make it non-executable. As there's no need for it.


bblimke:
Thanks! I don't even remember how it got there 👍

Issue #810, opened 2019-03-19
WebMockManticoreClient#response_object_for incorrectly invokes `call`


brianmcgue:
`WebMockManticoreClient#response_object_for` overrides [`Manticore::Client#response_object_for`](https://github.com/cheald/manticore/blob/79fbbc51fd5539479b4070eb66cd86de15b2ffb8/lib/manticore/client.rb#L440:L449).

`WebMockManticoreClient#response_object_for` invokes the `call` method on the response object, whereas `Manticore::Client#response_object_for` returns a response object without invoking the `call` method.

Implementations that depend on the behavior of the actual `Manticore::Client` may end up receiving an [`Already called`](https://github.com/cheald/manticore/blob/79fbbc51fd5539479b4070eb66cd86de15b2ffb8/lib/manticore/response.rb#L47) error.

brianmcgue:
Closed by https://github.com/bblimke/webmock/pull/811

PR #811, opened 2019-03-20
Use Manticore::Response#on_complete to invoke callbacks

brianmcgue:
Solves https://github.com/bblimke/webmock/issues/810


brianmcgue:
I'm happy to squash the commits if that will help.

I think the spec failure is the same as in master at the moment: https://travis-ci.org/bblimke/webmock

bblimke:
Thank you @brianmcgue

Issue #814, opened 2019-04-10
What is the correct usage of `net_connect_allowed?`


lucasuyezu:
The following code snippet explains the problem I'm having:

```
$ ruby -v
ruby 2.5.5p157 (2019-03-15 revision 67260) [x86_64-darwin18]
$ gem list webmock

*** LOCAL GEMS ***

webmock (3.5.1)
$ irb
irb(main):001:0> require 'webmock'
=> true
irb(main):002:0> WebMock.net_connect_allowed?
=> nil
irb(main):003:0> WebMock.disable_net_connect!(allow: "http://192.168.64.2:20031")
=> nil
irb(main):004:0> WebMock.net_connect_allowed?
Traceback (most recent call last):
        4: from /opt/rubies/2.5.5/bin/irb:11:in `<main>'
        3: from (irb):4
        2: from /Users/redacted/.gem/ruby/2.5.5/gems/webmock-3.5.1/lib/webmock/webmock.rb:63:in `net_connect_allowed?'
        1: from /Users/redacted/.gem/ruby/2.5.5/gems/webmock-3.5.1/lib/webmock/webmock.rb:75:in `net_connect_explicit_allowed?'
NoMethodError (undefined method `host' for nil:NilClass)
irb(main):005:0> ^C
irb(main):005:0>
```

Is this a bug, or is this just not the correct usage of `net_connect_allowed?`? Should I always call passing a URI string?

thatsmydoing:
I've also run into this. I think the behavior in your PR makes sense as there is no way to access `allow_net_connect` otherwise. Is there any reason you closed your PR?

bblimke:
@thatsmydoing thank you for reminding me of that. 

@lucasuyezu thank you for the issue and the PR. the branch was already deleted, therefore it wasn't possible to merge the PR and I have created a new commit.  https://github.com/bblimke/webmock/commit/c20bb68eb345fdc3ad4c7c25fe4971451585e88b

PR #815, opened 2019-04-10
Fix  with no arguments when allow is a non-localhost URI string

lucasuyezu:
Fixes https://github.com/bblimke/webmock/issues/814.

From the issue:

```
$ ruby -v
ruby 2.5.5p157 (2019-03-15 revision 67260) [x86_64-darwin18]
$ gem list webmock

*** LOCAL GEMS ***

webmock (3.5.1)
$ irb
irb(main):001:0> require 'webmock'
=> true
irb(main):002:0> WebMock.net_connect_allowed?
=> nil
irb(main):003:0> WebMock.disable_net_connect!(allow: "http://192.168.64.2:20031")
=> nil
irb(main):004:0> WebMock.net_connect_allowed?
Traceback (most recent call last):
        4: from /opt/rubies/2.5.5/bin/irb:11:in `<main>'
        3: from (irb):4
        2: from /Users/redacted/.gem/ruby/2.5.5/gems/webmock-3.5.1/lib/webmock/webmock.rb:63:in `net_connect_allowed?'
        1: from /Users/redacted/.gem/ruby/2.5.5/gems/webmock-3.5.1/lib/webmock/webmock.rb:75:in `net_connect_explicit_allowed?'
NoMethodError (undefined method `host' for nil:NilClass)
irb(main):005:0> ^C
irb(main):005:0>
```

### DIFF
--- a/lib/webmock/webmock.rb
+++ b/lib/webmock/webmock.rb
@@ -59,6 +59,8 @@ module WebMock
   end
 
   def self.net_connect_allowed?(uri = nil)
+    return Config.instance.allow_net_connect if uri.nil?
+
     if uri.is_a?(String)
       uri = WebMock::Util::URI.normalize_uri(uri)
     end
--- a/spec/unit/util/webmock_spec.rb
+++ b/spec/unit/util/webmock_spec.rb
@@ -1,0 +1,41 @@
+require 'spec_helper'
+
+module WebMock
+  describe ".net_connect_allowed?" do
+    context 'enabled globally' do
+      before do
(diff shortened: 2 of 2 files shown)


lucasuyezu:
There are build failures but they seem to be the same as the latest master. Don't know if they are related.

bblimke:
Thank you @lucasuyezu

lucasuyezu:
You're welcome! Glad to see it implemented.

bblimke:
This is now released as version 3.9.0

Issue #174, opened 2012-04-09
Matcher with a block


dnesteryuk:
Please, take a look at this line:

```
WebMock.should have_requested(:post, "www.example.com").with { |req| req.body == "abc" }
```

The code in the block will never be executed. I tried to raise an error there, tests passed.

I used:

```
Ruby 1.9.2
Webmock 1.8.6
Rspec 2.8.0
CurbFu 0.6.1
Curb 0.8.0
```

PR #820, opened 2019-05-22
Add a hint to the error message raised when `with` method is called w…

asok:
…ithout block

Related to #174


asok:
WDYT?

bblimke:
@asok thanks! That's a good idea. It happened to me so many times, not necessarily with WebMock.

tijmenb:
@asok I just got this excellent error message, and came here to thank you! Very helpful. 💯

asok:
So nice to get a simple, direct thank you :)

Issue #822, opened 2019-05-27
`hashdiff` Uses Non-Standard Constant


jfelchner:
Due to this issue: https://github.com/liufengyun/hashdiff/issues/45 `hashdiff` will be removing its current constant in the 1.0 release.  Due to webmock's popularity I thought it prudent to let you all know considering your gemspec does not currently have a pessimistic lock on that gem.

The upgrade path is extremely simple.  A find and replace in the project from `HashDiff` to `Hashdiff` will work exactly the same as before.  No require path changes are necessary.

Once `hashdiff` cuts `0.4.0`, I'd suggest cutting a webmock version with those constants renamed, and putting a `'>= 0.4.0'` on `hashdiff` in the gemspec.  Then, when we cut the `1.0`, the only change in webmock should be changing the version in the gemspec to `~> 1.0`.

jfelchner:
@bblimke I went ahead and created a PR for you. ;)

I'll let you know as soon as `hashdiff` `0.4.0` is released.

McRip:
`0.4.0` has been releasend. see https://rubygems.org/gems/hashdiff/versions/0.4.0.

jfelchner:
@McRip see the PR for more info: https://github.com/bblimke/webmock/pull/823#issuecomment-496723279

bblimke:
@jfelchner WebMock 3.6.0 has been released. Thank you for the heads-up before releasing hashdiff 0.4.0 and for PR. I very much appreciate that :)

jfelchner:
@bblimke no problem. :)  Just another heads up that, depending on the complaining about the warning, we'll be releasing `hashdiff` `1.0` which removes the shim at some point soon.  The `3.6.0` version of `webmock` should auto-upgrade to that and the warning your users are experiencing should simply go away.

teeparham:
I'm still seeing the warning with webmock 3.6.0 + hashdiff 0.4.0:

```
$ gem list hashdiff

*** LOCAL GEMS ***
hashdiff (0.4.0)

$ gem list webmock

*** LOCAL GEMS ***
webmock (3.6.0)

$ irb
> require "webmock"
The HashDiff constant used by this gem conflicts with another gem of a similar name.  As of version 1.0 the HashDiff constant will be completely removed and replaced by Hashdiff.  For more information see https://github.com/liufengyun/hashdiff/issues/45
```

bblimke:
@jfelchner when do you plan to release 1.0?

[...]

PR #823, opened 2019-05-28
Update `Hashdiff` Constant

jfelchner:
Why This Change Is Necessary
========================================================================

The `hashdiff` gem is changing its constant and thus we need to do the
same.

How These Changes Address the Issue
========================================================================

Update `HashDiff` references to `Hashdiff`.

Side Effects Caused By This Change
========================================================================

This will not remove the `hashdiff` warning error until `hashdiff`'s
`1.0` version is released, at which point a `bundle update` will remove
the warning.

------------------------------------------------------------------------
Actions:
  * References #822

### DIFF
--- a/lib/webmock/request_body_diff.rb
+++ b/lib/webmock/request_body_diff.rb
@@ -12,7 +12,7 @@ module WebMock
     def body_diff
       return {} unless request_signature_diffable? && request_stub_diffable?
 
-      HashDiff.diff(request_signature_body_hash, request_stub_body_hash)
+      Hashdiff.diff(request_signature_body_hash, request_stub_body_hash)
     end
 
     attr_reader :request_signature, :request_stub
--- a/webmock.gemspec
+++ b/webmock.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'addressable', '>= 2.3.6'
   s.add_dependency 'crack', '>= 0.3.2'
-  s.add_dependency 'hashdiff'
+  s.add_dependency 'hashdiff', ['>= 0.4.0', '< 2.0.0']
 
   unless RUBY_PLATFORM =~ /java/
     s.add_development_dependency 'patron',   '>= 0.4.18'


bblimke:
@jfelchner thank you. we can sync releases or the code can try to load Hashdiff constant and fallback to HashDiff is not present.

jfelchner:
@bblimke no problem! `hashdiff` `0.4.0` was just released and the PR should be safe to merge now.

jfelchner:
Once `0.4.0` is in the wild for a couple months, and your users have a chance to upgrade to latest, we'll release `1.0`.

jfelchner:
@bblimke because of webmock's popularity and the fact that the prior versions aren't pessimistically locked, we may want to make these same changes on the 1.x and 2.x branches and release new versions for them as well.

There are some popular gems that are locked to the 1.x and 2.x versions.

teeparham:
If you re-run the travis build, it should work, now that `0.4.0` is released.

jfelchner:
@teeparham Yep. At this point I'm just waiting for a reply from @bblimke

jodosha:
@jfelchner Thanks for fixing this annoying warning. 👍

jfelchner:
@bblimke can we get this merged?

jfelchner:
@jodosha strictly speaking, this won't remove the warning just yet.  But it will as soon as we release `hashdiff` `1.0` and you `bundle update`.

jfelchner:
@bblimke do you want me to create patches for v1 and v2 as well?

bblimke:
@jfelchner not sure it's worth supporting these old releases. people can always set versions of hashdiff in their Gemfiles. Though feel free to prepare these. I can release them as well.

I'll try to cut a new 3.x release tonight.

jfelchner:
@bblimke not a problem.  Up to you.  I just wanted to offer. :)

ndbroadbent:
For anyone else who wants to upgrade and remove the warning:

I upgraded to [webmock version `3.6.0`, which includes this change](https://github.com/bblimke/webmock/commit/1a33665f254cb15dfd3702ff58728320b7f8930f) (renames the constant from HashDiff to Hashdiff). I also [followed these instructions to upgrade to the hashdiff 1.0 beta](https://github.com/liufengyun/hashdiff/issues/45#issuecomment-499566400):

```
gem 'hashdiff', ['>= 1.0.0.beta1', '< 2.0.0']
```

Then run `bundle update hashdiff`. This makes sure you are using the latest versions and will get rid of the warning.

PR #830, opened 2019-07-21
stubbing io on stubbed socket

trusche:
Hi there!

I'm writing an app that calls some methods from the bundler gem programmatically. I don't want to spam the rubygems API, so I'm trying to use VCR and webmock to record my test requests. 

Now bundler (2.0.2) does some funky low-level stuff with Net::Http sockets that currently cause the following exception when used in conjunction with Webmock and VCR:

```
 NoMethodError:
       undefined method `io' for #<StubSocket:0x00007fde11d49508 @read_timeout=10>
```

Specifically, this is the result of a call to [Bundler::Definition#resolve_remotely](https://github.com/bundler/bundler/blob/master/lib/bundler/definition.rb#L154), and the immediate call triggering the exception is [this](https://github.com/bundler/bundler/blob/master/lib/bundler/vendor/net-http-persistent/lib/net/http/persistent.rb#L708).

This little patch fixes the issue by simply stubbing out the required object and method. I realize this is quite obscure and might not be worth merging. Let me know if it is please and I can try to come up with some code to reproduce this, at the least, and perhaps add a couple of tests to the PR.

Thanks!

### DIFF
--- a/lib/webmock/http_lib_adapters/net_http.rb
+++ b/lib/webmock/http_lib_adapters/net_http.rb
@@ -251,6 +251,13 @@ class StubSocket #:nodoc:
   def readuntil(*args)
   end
 
+  def io
+    @io ||= StubIO.new
+  end
+
+  class StubIO
+    def setsockopt(*args); end
+  end
 end
 
 module Net  #:nodoc: all
--- a/spec/acceptance/net_http/net_http_spec.rb
+++ b/spec/acceptance/net_http/net_http_spec.rb
@@ -201,6 +201,18 @@ describe "Net:HTTP" do
     expect(Net::HTTP.get_response(Addressable::URI.parse('http://www.example.com/hello?a=1')).body).to eq("abc")
   end
 
+  it "should support method calls on stubbed socket" do
+    WebMock.allow_net_connect!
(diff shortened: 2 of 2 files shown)


bblimke:
Hi @trusche

The fix looks good. Thank you for raising that and submitting the PR. It would be good to provide at least one spec to cover that. E.g. to make sure that nobody removes that code by mistake.

trusche:
Cool. I added a spec that does fail before and passes after my test - took me a bit to figure out how to construct the right stubbed request. Let me know if this makes sense, please.

trusche:
BTW I tried to troubleshoot that persistent Travis failure for ruby 2.6.2 (https://travis-ci.org/bblimke/webmock/jobs/562205529) but couldn't reproduce it locally (on OS X). One of those nasty only-on-CI issues? Got any leads I could pursue? I couldn't find an open issue for it.

joshcooper:
Running into the same issue. This PR should fix issue https://github.com/bblimke/webmock/issues/854

bblimke:
I'm sorry, I have overlooked this. Thank you for adding the test @trusche

bblimke:
This is now released in version 3.9.0. I'm sorry it took so long to release it.

Issue #827, opened 2019-07-02
Mock doesn't work if it matches `:80` or `:443`


thisismydesign:
```
require 'webmock'
require 'net/http'

include WebMock::API
WebMock.enable!

stub_request(:get, %r{example.com/user:1}).to_return(body: "{}")
Net::HTTP.get('https://example.com', '/user:1')
# => '{}'

stub_request(:get, "http://example.com/user:443").to_return(body: "{}")
Net::HTTP.get('http://example.com', '/user:443')
# => WebMock::NetConnectNotAllowedError

stub_request(:get, %r{example.com/user:443}).to_return(body: "{}")
Net::HTTP.get('https://example.com', '/user:443')
# => WebMock::NetConnectNotAllowedError

stub_request(:get, %r{example.com/user:443/likes}).to_return(body: "{}")
Net::HTTP.get('https://example.com', '/user:443/likes')
# => WebMock::NetConnectNotAllowedError

stub_request(:get, %r{example.com/user:80}).to_return(body: "{}")
Net::HTTP.get('https://example.com', '/user:80')
# => WebMock::NetConnectNotAllowedError

stub_request(:get, %r{example.com/user:801}).to_return(body: "{}")
Net::HTTP.get('https://example.com', '/user:801')
# => WebMock::NetConnectNotAllowedError

stub_request(:get, %r{example.com/user:4435}).to_return(body: "{}")
Net::HTTP.get('https://example.com', '/user:4435')
# => WebMock::NetConnectNotAllowedError
```

The error itself is also very misleading since you can see the correct stub registered (i.e. it doesn't print the correct URL which is being stubbed):
```
WebMock::NetConnectNotAllowedError (Real HTTP connections are disabled. Unregistered request: GET http://https//example.com:80/user:80 with headers {'Accept'=>'*/*', 'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'User-Agent'=>'Ruby'})

You can stub this request with the following snippet:

stub_request(:get, "http://https//example.com:80/user:80").
  with(
    headers: {
          'Accept'=>'*/*',
          'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3',
          'User-Agent'=>'Ruby'
    }).
  to_return(status: 200, body: "", headers: {})

registered request stubs:

stub_request(:get, "/example.com\/user:80/")
```

My guess is that this line is responsible:
[...]

PR #832, opened 2019-07-30
Fix overeager removal of default ports from urls when matching

fcheung:
:80 / :443 was removed no matter where it appeared in the url, ie
also in path or authentication sections. Fixed by using URI to remove
the port rather than regexps

See also #827


fcheung:
looks like the same failure against ruby 2.6.2 that other recent builds have had

bblimke:
@fcheung great, thank you!

PR #833, opened 2019-07-30
Fix a sporadic spec error

fcheung:
I observed builds randomly failing, mostly on linux with EPIPE errors.

I *think* this is webrick closing the connection before the client has sent the request. This change makes webrick wait for the first byte of the request before sending a response.

There was another failure that was because we weren’t passing through the exception:false flag to rbuf_fill causing it to raise exceptions the net http code wasn’t expecting.


bblimke:
Oh wow, this error was so annoying. Thank you!

PR #834, opened 2019-07-31
Fix to support HTTP/2 status lines

fabiob:
The regular expression applied to the HTTP status line only supports strings starting with "HTTP/x.x". However, HTTP/2 status lines won't use the dot.


bblimke:
Thank you @fabiob

PR #835, opened 2019-08-01
Fix EM HTTP request adapter's timeout class

tadassce:
As defined in a test, that's the expected error class:

https://github.com/igrigorik/em-http-request/blob/907f46713e8257e1c1dd348bc5da77adc8d38392/spec/client_spec.rb#L499


bblimke:
@tadassce looks good. thank you!

PR #839, opened 2019-08-21
Add support for Async::HTTP::Client

ayanko:
Add support for new asynchronous http library.
See https://github.com/socketry/async-http

### DIFF
--- a/spec/acceptance/async_http_client/async_http_client_spec_helper.rb
+++ b/spec/acceptance/async_http_client/async_http_client_spec_helper.rb
@@ -1,0 +1,73 @@
+module AsyncHttpClientSpecHelper
+  def http_request(method, url, options = {}, &block)
+    endpoint = Async::HTTP::Endpoint.parse(url)
+
+    path = endpoint.path
+    path = path + "?" + options[:query] if options[:query]
+
+    headers = (options[:headers] || {}).each_with_object([]) do |(k, v), o|
+      Array(v).each do |v|
+        o.push [k, v]
+      end
+    end
+    headers.push(
+      ['authorization', 'Basic ' + Base64.strict_encode64(options[:basic_auth].join(':'))]
+    ) if options[:basic_auth]
+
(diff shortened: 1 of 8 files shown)


bblimke:
@ayanko thank you for the PR. It will be great to support Async::HTTP in Webmock.

I'd like to make sure that the adapter is compatible with all WebMock usecases. Perhaps I have missed that, but I don't see `include_examples "with WebMock"` anywhere is the spec.

Passing `"with WebMock"` examples is required for all http client adapters to guarantee sure they are fully compatible with WebMock API.

I guess that some AsyncHttpRequestSpecHelper will be needed as well for "with WebMock" compatibility.

ayanko:
@bblimke Updated

* Added spec helper
*  Introduced`:no_content_length_header` option
* Added global shared examples

ayanko:
@bblimke 

`Async::HTTP::Client` deletes `Content-Length` header from server response so I added :no_content_length_header option to shared specs.


See https://github.com/socketry/protocol-http1/blob/master/lib/protocol/http1/connection.rb#L441 for details

spec/acceptance/async_http_client/async_http_client_spec_helper.rb
```diff
@@ -0,0 +1,73 @@
...
+          )
+        end
+      rescue Exception => e
+        e
+      end
+    end.wait
+  end
+
+  def client_timeout_exception_class
+    "WebMock timeout error"
```
bblimke: @ayanko shouldn't that be Async::TimeoutError?

ayanko: @bblimke Yeap. Updated.

ayanko:
@bblimke What else?

bblimke:
@ayanko I believe it's all good. Thank you.

bblimke:
Released as 3.7.0

PR #843, opened 2019-09-04
Do not try to gsub on incompatible encoding

rafaelfranca:
After #832, if the URL contains incompatible encoding characters the gsub to remove the protocol relative from the URL will fail. We should encode the string to binary mode before doing the gsub.

Got the tests from https://github.com/bblimke/webmock/pull/687 since it is the same issue so this doesn't regress anymore.

### DIFF
--- a/lib/webmock/util/uri.rb
+++ b/lib/webmock/util/uri.rb
@@ -51,7 +51,10 @@ module WebMock
           uris = uris_with_scheme_and_without(uris)
         end
 
-        uris.map {|uri| uri.to_s.gsub(/^\/\//,'') }.uniq
+        uris.map do |uri|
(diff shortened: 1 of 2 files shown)


bblimke:
Thank you @rafaelfranca . Could you please point me to the gsub that is failing before this fix?

rafaelfranca:
Of course, it is this one https://github.com/bblimke/webmock/blob/5b1754b3a6106cbfa58f9386b04e47c38b0f84a6/lib/webmock/util/uri.rb#L54.

I changed the original patch a little but in the first commit I only changed that gsub, so it is easier to see.

bblimke:
Thank you!

bblimke:
Released as 3.7.2

rafaelfranca:
Thank you!

PR #844, opened 2019-09-05
Add test for regression caused by #829

rafaelfranca:
I found a regression caused by #829 if we try to match an `Addressable::Template` using an ip and without any patch.

This is the failure I get:

```
  1) WebMock::RequestPattern when matching should match if uri Addressable::Template pattern matches request uri without a schema and a path
     Failure/Error: WebMock::Util::URI.variations_of_uri_as_strings(uri).any? { |u| normalized_template.match(u) }

     Addressable::URI::InvalidURIError:
       Invalid scheme format: 127.0.0.1
     # ./lib/webmock/request_pattern.rb:186:in `block in matches_with_variations?'
     # ./lib/webmock/request_pattern.rb:186:in `any?'
     # ./lib/webmock/request_pattern.rb:186:in `matches_with_variations?'
     # ./lib/webmock/request_pattern.rb:158:in `matches?'
     # ./lib/webmock/request_pattern.rb:37:in `matches?'
     # ./spec/unit/request_pattern_spec.rb:68:in `match'
     # ./spec/unit/request_pattern_spec.rb:138:in `block (3 levels) in <top (required)>'
```

Reverting the changes in #829 this test pass.

I'll try to also provide a fix but I'm opening a PR only with the test in case one of the maintainer know the best way to fix this case.

### DIFF
--- a/lib/webmock/util/uri.rb
+++ b/lib/webmock/util/uri.rb
@@ -47,7 +47,7 @@ module WebMock
 
         uris = uris_encoded_and_unencoded(uris)
 
-        if normalized_uri.scheme == "http" && !only_with_scheme
+        if normalized_uri.scheme == "http" && normalized_uri.host.start_with?(/[[:alpha:]]/) && !only_with_scheme
           uris = uris_with_scheme_and_without(uris)
         end
 
--- a/lib/webmock/util/uri.rb
+++ b/lib/webmock/util/uri.rb
@@ -47,7 +47,7 @@ module WebMock
 
         uris = uris_encoded_and_unencoded(uris)
 
-        if !only_with_scheme && normalized_uri.scheme == "http" && normalized_uri.host.start_with?(/[[:alpha:]]/)
+        if !only_with_scheme && normalized_uri.scheme == "http" && normalized_uri.host =~ /^[[:alpha:]]/
           uris = uris_with_scheme_and_without(uris)
         end
 
--- a/lib/webmock/util/uri.rb
+++ b/lib/webmock/util/uri.rb
@@ -47,7 +47,7 @@ module WebMock
 
         uris = uris_encoded_and_unencoded(uris)
 
-        if normalized_uri.scheme == "http" && normalized_uri.host.start_with?(/[[:alpha:]]/) && !only_with_scheme
+        if !only_with_scheme && normalized_uri.scheme == "http" && normalized_uri.host.start_with?(/[[:alpha:]]/)
           uris = uris_with_scheme_and_without(uris)
         end
 
--- a/spec/unit/request_pattern_spec.rb
+++ b/spec/unit/request_pattern_spec.rb
@@ -132,6 +132,12 @@ describe WebMock::RequestPattern do
       expect(WebMock::RequestPattern.new(:get, uri)).to match(signature)
     end
 
+    it "should match if uri Addressable::Template pattern matches request uri without a schema and a path " do
+      signature = WebMock::RequestSignature.new(:get, "127.0.0.1:3000")
+      uri = Addressable::Template.new("127.0.0.1:3000")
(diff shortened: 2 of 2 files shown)


rafaelfranca:
I can't think in a way to fix this problem without reverting that commit.

`localhost:3000` is a valid URI, but not in the way we think. `localhost` become the scheme here, not the host and `3000` is the path.

```
>> require 'addressable'
=> true
>> u = Addressable::URI.parse("localhost:3000")
=> #<Addressable::URI:0x3ff4560e7e20 URI:localhost:3000>
>> u.scheme
=> "localhost"
>> u.port
=> nil
>> u.path
=> "3000"
```

Because of that `127.0.0.1:3000` is not a valid URI because the schema needs to start with a letter.

When #8829 was made this method started to parse invalid URIs.

I think the way forward is to revert that change and see if there is another way to make domain with TLD match against URIs with schema.

rafaelfranca:
I added a proposed fix. Let me know what you think.

bblimke:
@rafaelfranca is this purpose of this fix to differentiate IP address from a hostname? how about https://9292.nl/ ?

rafaelfranca:
No. The purpose of this fix is to prevent webmock to try use `9292.nl` as a valid schema.

In a URIs like `foo:80` (or `9292.nl:80`) the `foo` is not a host (or IP), it is a schema. As you can see here:

```ruby
>> u = Addressable::URI.parse("localhost:3000")
=> #<Addressable::URI:0x3ff4560e7e20 URI:localhost:3000>
>> u.scheme
=> "localhost"
>> u.port
=> nil
>> u.path
=> "3000"
```

All URI parsers don't allow schemas that start with numbers. So what this PR is trying to do is prevent webmock to generate an URI like that when the host starts with a number.

Before #829, this problem didn't exist because `variations_of_uri_as_strings` was being called with `only_with_scheme: true`. This means that `variations_of_uri_as_strings` never returned URI where the schema started with numbers since the schema was always `http(s)`.

For hosts like `9292.nl`, things will work as they worked before #829.

bblimke:
@rafaelfranca ok, but in that case, if it only affects the templates, then perhaps the check for numbers should be made inside the pattern, where `variations_of_uri_as_strings` is invoked, and only_with_scheme should be passed.

Alternatively we could assume that URI without scheme and starting with numbers are not valid,
and then completely remove `only_with_scheme` argument (after your change it's not used anywhere), but this somehow breaks backwards compatibility, since really these are still valid variations of the url, just not accepted by Addressable::Template#match.

rafaelfranca:
I'm not sure the problem is with the pattern. Yes, it explodes when we are trying to match the pattern with the URI, but the pattern is correct, what is not is the URI. But I agree we can check for that in `matches_with_variations` and pass `only_with_scheme` in case the URI will become invalid without a schema. I'll work on that

casperisfine:
👋 is there interest in this PR? Because it's still a problem on the last published version.

alain-andre:
> wave is there interest in this PR? Because it's still a problem on the last published version.

When looking on the Internet, we are a lot having the problem, I noticed it in this [PR](https://github.com/bblimke/webmock/pull/758) and our team is now using this PR.

alan-pie:
@rafaelfranca @bblimke What is needed to move this forward?
https://github.com/Pie-Systems/webmock/tree/alan/master-with-address-fix has this fix updated with the merge conflict resolved

bblimke:
@alan-pie I believe this will be fixed when the following PR is merged https://github.com/bblimke/webmock/pull/890
It seems to be a more generic solution that covers many more possible error cases.

bblimke:
@rafaelfranca thank you for the PR. I have applied the following, more generic solution instead. https://github.com/bblimke/webmock/commit/72eee63f8a48a26bdfcbf69b9b75ccebd3b29e68

Your test passes.

bblimke:
The fix has now been released as version 3.9.0

PR #845, opened 2019-09-11
Allow passing an output buffer to HTTP::Response::Body#readpartial

georgeclaghorn:
Unmocked:

```
> response = HTTP.get "http://example.com/foo"
> response.body.readpartial(1024, "")
"Hello world!"
```

Mocked:

```
> response = HTTP.get "http://example.com/foo"
> response.body.readpartial(1024, "")
ArgumentError: wrong number of arguments (given 2, expected 0..1)
```

`HTTP::Response::Body` delegates `#readpartial` to the underlying connection. For an unmocked request, that’s an `HTTP::Connection`, which accepts the conventional second argument to `#readpartial`. For a mocked request, it’s an `HTTP::Response::Streamer`, which previously *didn’t* accept a second argument.

### DIFF
--- a/lib/webmock/http_lib_adapters/http_rb/streamer.rb
+++ b/lib/webmock/http_lib_adapters/http_rb/streamer.rb
@@ -5,7 +5,7 @@ module HTTP
         @io = StringIO.new str
       end
 
-      def readpartial(size = nil)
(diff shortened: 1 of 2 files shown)


bblimke:
Thank you @georgeclaghorn !

PR #847, opened 2019-09-13
Invoke manticore failure handler on stubbed timeout

alexJunger:
Stubbing timeouts with Manticore and Faraday currently does not work correctly.
Manticore's faraday adapter registers an on_failure handler which reraises `Manticore::ConnectTimeout` as `Faraday::TimeoutError`.

`ManticoreAdapter` does not invoke the on_failure handler if the mock should timeout, but instead just raises the timeout exception. This matches Manticore's default failure handler behaviour, but does not match what happens for other handlers (as in this case the faraday adapter's handler).


bblimke:
Thank you @alexJunger .

Issue #703, opened 2017-05-30
Same response from different request


dinhhuydh:
I currently use the Webmock `3.0.1` for `vcr` to stub external request. The issue is that the 2 different url requests return the same response so that the test is failed. 
When `vcr` is not used, the test is passed. The requests are not in the same spec/scenario.
As i look at the code, `Webmock` implements class called `Curl::WebMockCurlEasy` which inherited the `Curl::Easy`. Although it does not override the `reset` method, somehow it does not clear out the `@body_str` of the previous request. The `reset` of `Curl::Easy` does clear out the variable. One notice is that the `Curl::Easy` instance is the same in the two requests. 
Because the `reset` in the library is in C code so i can't go further.

If i apply the patch to clear out the `body_str`, it works well.
```
class Curl::WebMockCurlEasy
  def reset
    instance_variable_set('@body_str', nil)
    super
  end
end
```

bblimke:
@dinhhuydh thank you for submitting the issue and providing a solution.

Would you fancy providing a pull request with that solution implemented?

PR #848, opened 2019-09-15
Reset memoized response fields in Curb adapter

heretge:
Exactly the same issue as in https://github.com/bblimke/webmock/issues/703 bites again!

I was running into a problem with VCR recording the same headers for a series of different Curl requests. Turns out it was due to partial `reset` method implementation, which only cleans up `body_str`, although `Curl::WebMockCurlEasy` memoizes a lot more things, including `header_str` for my case.


bblimke:
Great fix. Thank you @heretge

PR #860, opened 2019-11-22
Fixes ruby 2.7 warnings and other noisy test output

zenspider:


bblimke:
@zenspider I appreciate the kwargs changes, but why have you changed all dependencies in gemspec to "~>"? The ">=" is by design as WebMock needs to be compatible with multiple minor or even major versions.

zenspider:
@bblimke to shut up rubygems warnings. IIRC I was getting them every time I ran via rake or something.

Unbound dependencies are the root of many evils. I think if you want to be intentional about it, then you should do that explicitly with the one or two things you really want to have wide ranging compatibility with, but you should still clamp down on the others. Minitest for example, should be `"~> 5.0"` for sure. I _will_ break your shit when I put out 6.0.

Feel free to reject that commit. I can rebase it out... but do consider my advice.

zenspider:
P.S. preview2 should bump to preview3... The final 2.7.0 has also been released but docker images aren't out yet. I don't know if travis uses their own images or not.

bblimke:
> Unbound dependencies are the root of many evils.

I agree 100% :)

> I think if you want to be intentional about it, then you should do that explicitly with the one or two things you really want to have wide ranging compatibility with, but you should still clamp down on the others. 

I really want WebMock to be be explicitly compatible with all dependencies. It's of course impossible or would be too much headache to try to be compatible with every single version, therefore I try to focus on making WebMock compatible with latest versions, while trying to keep WebMock compatible with old versions as much as possible as well, but only if it's not too much headache.

Thank you for the heads up about minitest 6.0! :)
The idea behind keeping unbound dependencies in WebMock, is that when WebMock becomes incompatible with the latest version of some library, Travis build will break or the build will break at least for someone, who will raise an issue. Otherwise people are free to use the latest versions of all http clients, Addressable, Minitest or Rspec and there is no need to restrict these dependencies. 

That's how WebMock has worked for the last 10 years and it worked well enough. I believe it worked because backward incompatible changes rarely happen to any of the http client gems.

I believe it makes sense to keep http lib adapters dependencies unbound. Major version upgrade doesn't necessarily mean that WebMock becomes incompatible and there is nobody who would track all the recent versions of http client libs to see if the latest changes are still compatible with WebMock. Almost always when there is incompatibility, things will break.

I'm happy to set the dependency on minitest to "~> 5.0" or Rspec, but I'd rather still keep unbound dependencies on all http clients.

koic:
FYI, [Ruby 2.7.0](https://www.ruby-lang.org/en/news/2019/12/25/ruby-2-7-0-released/) matrix available on Travis CI. 
http://rubies.travis-ci.org/

bblimke:
Thank you @zenspider I have cherry-picked two commits and released version 3.8.0, but I still need to consider whether to change the dependencies.

PR #868, opened 2020-01-06
Fix options handling.

### LINKED ISSUE #867: undefined method `client' for {}:Hash
Something strange is going on in the async-http monkey patching.

```
     NoMethodError:
       undefined method `client' for {}:Hash
     # ./vendor/bundle/ruby/2.4.0/gems/async-http-0.50.0/lib/async/http/client.rb:140:in `block in make_pool'
     # ./vendor/bundle/ruby/2.4.0/gems/async-pool-0.2.0/lib/async/pool/controller.rb:181:in `create_resource'
     # ./vendor/bundle/ruby/2.4.0/gems/async-pool-0.2.0/lib/async/pool/controller.rb:210:in `block in available_resource'
     # ./vendor/bundle/ruby/2.4.0/gems/async-1.24.1/lib/async/semaphore.rb:78:in `acquire'
     # ./vendor/bundle/ruby/2.4.0/gems/async-pool-0.2.0/lib/async/pool/controller.rb:191:in `available_resource'
     # ./vendor/bundle/ruby/2.4.0/gems/async-pool-0.2.0/lib/async/pool/controller.rb:166:in `wait_for_resource'
     # ./vendor/bundle/ruby/2.4.0/gems/async-pool-0.2.0/lib/async/pool/controller.rb:76:in `acquire'
     # ./vendor/bundle/ruby/2.4.0/gems/async-http-0.50.0/lib/async/http/client.rb:104:in `call'
     # ./vendor/bundle/ruby/2.4.0/gems/webmock-3.7.6/lib/webmock/http_lib_adapters/async_http_client_adapter.rb:63:in `call'
```

ioquatix:
Fixes #867

### DIFF
--- a/lib/webmock/http_lib_adapters/async_http_client_adapter.rb
+++ b/lib/webmock/http_lib_adapters/async_http_client_adapter.rb
@@ -36,7 +36,7 @@ if defined?(Async::HTTP)
           protocol = endpoint.protocol,
           scheme = endpoint.scheme,
           authority = endpoint.authority,
-          options = {}
+          **options
         )
           webmock_endpoint = WebMockEndpoint.new(scheme, authority, protocol)
 


bblimke:
Thank you @ioquatix. Is there any easy way to add spec example for that. How would async-http invocation that takes advantage of this look like?

ioquatix:
There isn't really a way to add a spec, it's simply broken and wrong. I'm pretty sure if you update to the latest version of async-http and use `Async::Internet` it will be broken. So you could add a spec to see if it's working normally.

Using `options = {}` is no longer good practice and you should replace with `**options` everywhere.

bblimke:
Ok, thanks! :)

Issue #552, opened 2015-12-11
Typhoeus::Response#effective_url should be a string, Addressable::URI returned


erwanlr:
Related to #540 

As mentioned in the Typhoeus' doc, the effective_url method should return a string: https://github.com/typhoeus/typhoeus/blob/de61860fecec32b58e150e63219d1a4062a77855/lib/typhoeus/response/informations.rb#L181

Currently in webmock, an Addressable::URI object is returned

The fix would be to replace https://github.com/bblimke/webmock/blob/bfc8206a8c728c5f93dca7ee13a83f4e7ebc79e3/lib/webmock/http_lib_adapters/typhoeus_hydra_adapter.rb#L101 by ```:effective_url => request_signature.uri.to_s```

erwanlr:
Furthermore, the default port (either 80 or 443 depending on the http/s used) is returned, which screw up specs a bit :/

tbrammar:
Implemented a simple PR here to fix this: https://github.com/bblimke/webmock/pull/869

PR #869, opened 2020-01-07
Return effective_url as string (without port details)

tbrammar:
This is an attempted fix for this issue https://github.com/bblimke/webmock/issues/552


bblimke:
Thank you @tbrammar. Looks good but I believe only the default port should be omitted?

tbrammar:
@bblimke good catch 👍 - have updated to only remove standard ports

PR #871, opened 2020-01-09
Ruby 2.7: Use keyword args double-asterisk in `WebMockNetBufferedIO`

brandur:
Webmock is currently showing a warning on Ruby 2.7 related to the use of
a single asterisk for passing keyword args to `BufferedIO` via the use
of an argument-less `super`.

Here's the warning:

```
/Users/brandur/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/webmock-3.7.6/lib/webmock/http_lib_adapters/net_http.rb:271: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/brandur/.rbenv/versions/2.7.0/lib/ruby/2.7.0/net/protocol.rb:114: warning: The called method `initialize' is defined here
```

And here's the method in question:

``` ruby
  class WebMockNetBufferedIO < BufferedIO
    def initialize(io, **args)
      io = case io
      when Socket, OpenSSL::SSL::SSLSocket, IO
        io
      when StringIO
        PatchedStringIO.new(io.string)
      when String
        PatchedStringIO.new(io)
      end
      raise "Unable to create local socket" unless io

      super
    end
```

Taking a look at the definition of `BufferedIO` [1], it seems to purely
take keyword args after the first `io` argument:

``` ruby
initialize(io, read_timeout: 60, write_timeout: 60, continue_timeout: nil, debug_output: nil) ⇒ BufferedIO
```

So I believe it's safe here just to switch to a double asterisk from a
single and call it a day.

[1] https://www.rubydoc.info/stdlib/net/Net/BufferedIO

### DIFF
--- a/lib/webmock/http_lib_adapters/net_http.rb
+++ b/lib/webmock/http_lib_adapters/net_http.rb
@@ -268,7 +268,14 @@ module Net  #:nodoc: all
       end
       raise "Unable to create local socket" unless io
 
-      super
+      # Prior to 2.4.0 `BufferedIO` only takes a single argument (`io`) with no
+      # options. Here we pass through our full set of arguments only if we're
+      # on 2.4.0 or later, and use a simplified invocation otherwise.
+      if RUBY_VERSION >= '2.4.0'
+        super
+      else
+        super(io)
+      end
     end
 
     if RUBY_VERSION >= '2.6.0'
--- a/lib/webmock/http_lib_adapters/net_http.rb
+++ b/lib/webmock/http_lib_adapters/net_http.rb
@@ -257,7 +257,7 @@ end
 module Net  #:nodoc: all
 
   class WebMockNetBufferedIO < BufferedIO
-    def initialize(io, *args)
+    def initialize(io, **args)
       io = case io
       when Socket, OpenSSL::SSL::SSLSocket, IO
         io


brandur:
(Looks like some versions of Ruby don't like this — taking a look.)

brandur:
Okay, the problem was that prior to 2.4.0, `BufferedIO.new` only took a single option:

``` ruby
  class BufferedIO   #:nodoc: internal use only
    def initialize(io)
```

I fixed the patch for older versions by introducing a conditional based on `RUBY_VERSION`, which does the trick:

``` ruby
      # Prior to 2.4.0 `BufferedIO` only takes a single argument (`io`) with no
      # options. Here we pass through our full set of arguments only if we're
      # on 2.4.0 or later, and use a simplified invocation otherwise.
      if RUBY_VERSION >= '2.4.0'
        super
      else
        super(io)
      end
```

That's of course not quite as clean though, and introduces a slight problem where if `WebMockNetBufferedIO.new` is invoked with an option on an old version of Ruby then that option is silently ignored. That's probably okay given that Ruby 2.3 will have been EOLed for two years this coming March, so use is probably low.

One additional change I could make if desired is add an `ArgumentError` in case the user passes options on Ruby < 2.4:

``` ruby
diff --git lib/webmock/http_lib_adapters/net_http.rb lib/webmock/http_lib_adapters/net_http.rb
index 3fb4167..d767a18 100644
--- lib/webmock/http_lib_adapters/net_http.rb
+++ lib/webmock/http_lib_adapters/net_http.rb
@@ -274,6 +274,10 @@ module Net  #:nodoc: all
       if RUBY_VERSION >= '2.4.0'
         super
       else
+        unless args.empty?
+          raise ArgumentError, "BufferedIO options are not supported prior to Ruby 2.4"
+        end
+
         super(io)
       end
     end
```

It's probably not necessary though.

bblimke:
Thank you @brandur !

PR #873, opened 2020-01-29
Replace rbuf_fill modification with duplication of core library

kipply:
Why I think this is better than the previous work around made in https://github.com/bblimke/webmock/commit/1cc36e5d9a64f081cf83683752ceea6f04bf4ff5

- Maintenance complexity is likely lowered. Previous fix was susceptible to breaking whenever `rbuf_fill` was changed. This has the same concern in different ways though clearly references the code and is therefore easier to update. 
 - Avoids the overhead cost of creating tracepoint, generating binding and disable


bblimke:
@kipply Thank you for the PR and for the arguments! I agree on both points.

PR #876, opened 2020-02-11
Support correct encoding parameter for HTTP.rb 2.x and earlier

### LINKED ISSUE #875: Older HTTP.rb Versions Broken with 3.8.1
#874 passes through the body encoding to HTTP.rb, but uses keyword arguments which weren't used until v3. Based on [this version check](https://github.com/bblimke/webmock/blob/v3.8.1/lib/webmock/http_lib_adapters/http_rb_adapter.rb#L8), 0.6.0+ should be supported.

I can either submit a PR to bump that version to `3.0.0`, or use position or keyword arguments based on the gem's version, which would you prefer?

This is how the error manifests:

```
TypeError:
  no implicit conversion of Hash into String
# http-2.2.2/lib/http/response/body.rb:47:in `find'
# http-2.2.2/lib/http/response/body.rb:47:in `to_s'
# http-2.2.2/lib/http/response.rb:91:in `flush'
```

drcapulet:
Fixes #875.

### DIFF
--- a/lib/webmock/http_lib_adapters/http_rb/response.rb
+++ b/lib/webmock/http_lib_adapters/http_rb/response.rb
@@ -14,8 +14,18 @@ module HTTP
       def from_webmock(webmock_response, request_signature = nil)
         status  = Status.new(webmock_response.status.first)
         headers = webmock_response.headers || {}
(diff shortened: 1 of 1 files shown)


bblimke:
Thank you @drcapulet !

PR #878, opened 2020-03-19
Allow using a "callable" (like a proc) as URI pattern 

jhawthorn:
This PR allows providing any object which responds to `#call` as a URI pattern. This allows a lot more flexibility and accuracy than using a regex. One of the incentives for making this change is also that specifying a regex will call `variations_of_uri_as_strings` which is very slow. It's also much easier to stub all requests

The first commit is just a refactor of URIPattern and its subclasses, which makes introducing the new subclass as simple as possible.

This also changes `register_global_stub` to use `->(uri) {  true }` instead of `/.*/`, which is many times faster since it doesn't have to expand the URL patterns to perform the match.

### DIFF
--- a/lib/webmock/request_pattern.rb
+++ b/lib/webmock/request_pattern.rb
@@ -131,38 +131,36 @@ module WebMock
       end
     end
 
+    def matches?(uri)
+      pattern_matches?(uri) && query_params_matches?(uri)
+    end
+
     def to_s
-      str = @pattern.inspect
+      str = pattern_inspect
       str += " with query params #{@query_params.inspect}" if @query_params
       str
+    end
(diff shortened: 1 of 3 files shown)


bblimke:
@jhawthorn Thank you. It looks very useful!

My only concern is how the stub will be printed in a list of registered stubs,
since inspecting lambda produces something like `"#<Proc:0x00007fd8357af3e0@(irb):1 (lambda)>"`, but I'm not sure there is a good way to solve that, so let's ignore it and focus on advantages.

bblimke:
This is now released in version 3.9.0. I'm sorry it took so long to release it.

PR #880, opened 2020-04-02
CI: Use new Ruby, separate out JRuby

olleolleolle:
This PR updates the CI matrix to use latest releases. **Update**: too early for rvm! Seems these named versions from the https://www.ruby-lang.org/en/news/ from today are not yet available.

- Use different JDKs for 9.1 and 9.2+


bblimke:
Thank you @olleolleolle !

PR #882, opened 2020-04-17
Add WebMock.allow_net_connect to append to allowed

xjunior:
Add the ability to append to allowed hosts instead of completely replace it

### DIFF
--- a/lib/webmock/webmock.rb
+++ b/lib/webmock/webmock.rb
@@ -58,6 +58,10 @@ module WebMock
     Config.instance.allow.push *allowed
   end
 
+  def self.clear_allowed!
+    Config.instance.allow = []
+  end
+
   class << self
     alias :enable_net_connect!   :allow_net_connect!
     alias :disallow_net_connect! :disable_net_connect!
--- a/lib/webmock/webmock.rb
+++ b/lib/webmock/webmock.rb
@@ -52,7 +52,7 @@ module WebMock
     Config.instance.allow = Array(options[:allow])
     Config.instance.net_http_connect_on_start = options[:net_http_connect_on_start]
   end
-  
+
   def self.allow!(*allowed)
     Config.instance.allow ||= []
     Config.instance.allow.push *allowed
--- a/lib/webmock/webmock.rb
+++ b/lib/webmock/webmock.rb
@@ -52,6 +52,11 @@ module WebMock
     Config.instance.allow = options[:allow]
     Config.instance.net_http_connect_on_start = options[:net_http_connect_on_start]
   end
+  
+  def self.allow!(*allowed)
+    Config.instance.allow ||= []
+    Config.instance.allow.push *allowed
+  end
 
   class << self
     alias :enable_net_connect!   :allow_net_connect!
--- a/lib/webmock/webmock.rb
+++ b/lib/webmock/webmock.rb
@@ -49,7 +49,7 @@ module WebMock
   def self.disable_net_connect!(options = {})
     Config.instance.allow_net_connect = false
     Config.instance.allow_localhost = options[:allow_localhost]
-    Config.instance.allow = options[:allow]
+    Config.instance.allow = Array(options[:allow])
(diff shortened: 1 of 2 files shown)


bblimke:
@xjunior thanks for submitting the pull request. Am I missing something or does your change the previous definition of the method and changes the previous behaviour?

xjunior:
@bblimke my bad. I didn't notice I was replacing an existing method. I renamed it to `allow!` and also added `clear_allowed!`. Also, I changed the behavior of `disable_net_connect!` to always set the `allow`ed as an array.

bblimke:
@xjunior thanks!

I have been thinking about this feature and I have some concerns.

I'm not sure about the method names `allow!` and `clear_allowed` are clear enough.
`allow!` doesn't explain what is actually being allowed
`clear_allowed` doesn't explain what is being cleared or what was allowed.

`Config.instance.allow` is only used when combined with disable_net_connect!.
If `disable_net_connect` has not been invoked, then `Config.instance.allow` has no effect.
In that case I wonder what is the usecase for these two new methods.
People will call `allow!` and they will expect it to have effect, but without `disable_net_connect!` it will not have any effect. With that in mind, having `allow` only available as an option to `disable_net_connect!` is more logical.

I have checked your added specs, but I don't see it as a reasonable usecase
to call `allow!` separately right after `disable_net_connect!`

The idea to add these methods to WebMock is that people are missing these methods
and they will start using them. With that in mind, I'd like to understand what specific
issue are these new methods solving and in what kind of usecase.

xjunior:
Hey, @bblimke. My use-case is simple: I want to have `rspec` extensions appending to allowed hosts, and these hosts cleared between tests. What If we make `disable_net_connect!` have that behavior instead? Also, make `allow_net_connect!` clear the allowed?

bblimke:
> What If we make disable_net_connect! have that behavior instead?

`:allow` as an option won't be clear if we start appending instead of resetting allowed hosts.
I don't think it should append.

Since you want to to clear the allowed hosts between tests, the current behaviour sounds more useful.

> make allow_net_connect! clear the allowed?

allowed only is used when disable_net_connect has been called. Otherwise it's unused. When you invoke disable_net_connect again, allowed will be reset.

> I want to have rspec extensions appending to allowed hosts, and these hosts cleared between tests. 

This is not going to change as too many setups rely on current behaviour and it would be a breaking change. It's something to consider for version 4.

You can always simply call disable_net_connect in before callback in your rspec helper.

bblimke:
@xjunior thank you for the reply.

I'm happy to accept the additions to the API, but I don't think their naming is intuitive enough.

`allow` is not fortunately names as a config variable, but it's not exposed in the API.

I'm not sure people will know what `allow!` or what `clear_allowed!` means. 

People could think that `allow!` allows something, but it only appends to `allowed` array,
and only takes effect when disable_net_connect has been called first.

I don't think `disable_net_connect` is very intuitive either, but that's another topic :)

Perhaps `append_to_allowed!`, but what is `allowed` is not clear.

bblimke:
@xjunior thank you. I'm still thinking about the names.
That looks more readable, to you and to me, and to all people know what `allowed` is :)
I believe currently to understand what `allowed` is, one needs to look inside WebMock code,
don't you think?
The value of this change is to provide an API to people who don't know already they can simply
append to `Config.instance.allow`

xjunior:
I actually agree. What I would like to have is all of the same behavior of disable net connect but appending to the allowed. What do you think?

bblimke:
@xjunior sure, I understand. The current `disable_net_connect!` doesn't expect the user to think about `allow` as some global variable. To API user, it's just an option passed to `disable_net_connect!`. Whenever `disable_net_connect!` is called, new option can be passed.

If we add option to allow `appending to the allowed`, then we are forcing API users to think about the fact there is some global `allowed` that can be appended and they need to control, what it was before anything is appended.

xjunior:
I get it, @bblimke. Thanks!

PR #888, opened 2020-05-28
Support httprb 5.x.x

hieuk09:
`httprb` 5 requires `request` in response, so it breaks webmock integration. 

I would like to write tests for this PR, but I'm don't have much idea; could you give me some direction?


dollerbill:
@hieuk09 any reason you chose to close this PR?

hieuk09:
@dollerbill httprb 5 is still in beta, so I think I'll go back to this PR when it releases official version.

Issue #889, opened 2020-06-09
delete invalid uri like `127.0.0.1:3000/1234`


guppy0356:
### Versions:

webmock: 3.8.3
rspec: 3.5.4

### Problem

ip address host which has no ports doesn't match request uri.

```
it "should match if Addressable::Template pattern that has ip address host without port matches request uri" do
  signature = WebMock::RequestSignature.new(:get, "127.0.0.1/1234")
  uri = Addressable::Template.new("127.0.0.1/{id}")
  expect(WebMock::RequestPattern.new(:get, uri)).to match(signature)
end
```

### Workaround

Delete invalid string in the results of `#matches_with_variations?`.

PR #890, opened 2020-06-09
Fix InvalidURIError with `ip` uri

guppy0356:
## problem

It works with most cases without uri scheme, but it raises an error with some cases like 127.0.0.1.

I've found that it causes the error when `Capybara.server_host` is `0.0.0.0`.

## solution

The difference from https://github.com/bblimke/webmock/pull/758 is 

## issue

fixes: https://github.com/bblimke/webmock/issues/889

- delete invalid uri

### DIFF
--- a/lib/webmock/request_pattern.rb
+++ b/lib/webmock/request_pattern.rb
@@ -203,7 +203,15 @@ module WebMock
         rescue Addressable::URI::InvalidURIError
           Addressable::Template.new(@pattern.pattern)
         end
-      WebMock::Util::URI.variations_of_uri_as_strings(uri).any? { |u| template.match(u) }
+      uris = WebMock::Util::URI.variations_of_uri_as_strings(uri)
+      uris.select! { |uri| valid_uri?(uri) }
+      uris.any? { |u| template.match(u) }
+    end
+
+    def valid_uri?(uri)
+      !!Addressable::URI.parse(uri)
+    rescue Addressable::URI::InvalidURIError => e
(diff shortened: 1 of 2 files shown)


lib/webmock/request_pattern.rb
```diff
@@ -203,7 +203,15 @@ def matches_with_variations?(uri)
         rescue Addressable::URI::InvalidURIError
           Addressable::Template.new(@pattern.pattern)
         end
-      WebMock::Util::URI.variations_of_uri_as_strings(uri).any? { |u| template.match(u) }
+      uris = WebMock::Util::URI.variations_of_uri_as_strings(uri)
+      uris.select! { |uri| valid_uri?(uri) }
```
bblimke, line 207: @guppy0356 do you think it's worth moving `valid_uri?(uri)` inside the any block? it would be faster not to iterate over all uris, but only until the first valid and matching.

bblimke:
Thank you. That looks like a good solution. I added one comment to the code.

bblimke:
@guppy0356 Thank you. I have applied similar solution, but more optimised as it doesn't require expensive parsing of each uri variation https://github.com/bblimke/webmock/commit/72eee63f8a48a26bdfcbf69b9b75ccebd3b29e68

bblimke:
The fix has now been released as version 3.9.0

guppy0356:
@bblimke 
Sorry for the late reply. I will update our product Gemfie.

PR #898, opened 2020-07-31
Include documentation RegEx hash_including matcher

ChrisBr:
I just learned about this really cool feature that you can match agains a RegEx in `hash_including` so I added some documentation for it.

```ruby
stub_request(:post, "www.example.com").
  with(body: hash_including(id: /\d/))

RestClient.post('www.example.com', "id=1",
:content_type => 'application/x-www-form-urlencoded')    # ===> Success
```


bblimke:
@ChrisBr this only works if we are inside RSpec. WebMock has own limited implementation of hash_including in case hash_including is not already available. Perhaps it's better to mention that in the documentation, that people can take advantage of all RSpec's hash_including goodies as long as it's Rspec.

PR #901, opened 2020-09-08
Fix async http adapter

### LINKED ISSUE #858: Async:HTTP never returns
Stubbing an `Async::HTTP::Client` or `Async::HTTP::Internet` does not return any response for me, but hangs for ever.

Using async-http (0.48.2), webmock (3.7.6), rspec-rails (3.9.0), rails (6.0.1)

Here's some sample code to illustrate:

```ruby
RSpec.describe "Async::HTTP" do
  before { WebMock.disable_net_connect! }

  context "with Internet" do
    class InternetClient
      def initialize(http)
        @http = http
        @host = 'http://example.com'
        @headers =  Protocol::HTTP::Headers[{
          content_type: 'application/json',
          accept: 'application/json'
        }]
      end
      attr_reader :http, :host, :headers

      def post_stuff(body)
        response = http.post("#{host}/stuff", headers, [body])

        if response.success?
          [true, response.read]
        else
          [false, response.read, response.status]
        end
      end
    end

    subject { InternetClient.new(Async::HTTP::Internet.new) }

    specify do
      stub_request(:post, "http://example.com/stuff")
        .with(body: '"request body"')
        .to_return(body: '"response body"')

      expect(subject.post_stuff('"request body"')).to eq([true, '"response body"'])
    end
  end

  context "with Client" do
    class HttpClient
      def initialize(http)
        @http = http
        @headers = {
          content_type: 'application/json',
          accept: 'application/json'
        }
      end
      attr_reader :http, :headers

      def post_stuff(body)
        response = http.post("/stuff", headers, [body])

        if response.success?
          [true, response.read]
        else
          [false, response.read, response.status]
        end
      end
    end

    subject { HttpClient.new(Async::HTTP::Client.new(Async::HTTP::Endpoint.parse('http://example.com'))) }

    specify do
      stub_request(:post, "http://example.com/stuff")
        .with(body: '"request body"')
        .to_return(body: '"response body"')

      expect(subject.post_stuff('"request body"')).to eq([true, '"response body"'])
    end
  end
end
```

bruno-:
Problem: the server in Async::HTTP::WebMockEndpoint#accept_socket was
hanging forever after the rspec example finished.
This is normal as Async::HTTP::Server (or any server) is never supposed
to end. It just continues waiting for new requests which, in this case,
never come.

Solution: use a 'transient' async task when starting a server.
Transient tasks do not block an exiting async reactor.
So, in this case a transient server will not hang forever.

Fixes #858 

@ioquatix we talked about making this a draft PR so we can continue talking about a proper webmock <-> async-http solution. Since this fix is minimal I propose going for a merge.
We can continue thinking about a better approach here https://github.com/socketry/async-http/issues/35

### DIFF
--- a/lib/webmock/http_lib_adapters/async_http_client_adapter.rb
+++ b/lib/webmock/http_lib_adapters/async_http_client_adapter.rb
@@ -136,7 +136,7 @@ if defined?(Async::HTTP)
 
         def connect
           server_socket, client_socket = create_connected_sockets
-          Async do
+          Async(transient: true) do
             accept_socket(server_socket)
           end
           client_socket


ioquatix (approved):

bruno-:
@ioquatix I just saw your latest comment https://github.com/bblimke/webmock/issues/858#issuecomment-688687027

Here is the "expanded" code from this PR:

```ruby
Async(transient: true) do  # <= This task IS the server's task
  server = Async::HTTP::Server.new(WebMockApplication, self)
  server.accept(socket, socket.remote_address)
end
```

Is the `ensure` block with `task&.stop` necessary? If yes, we'd need to insert another `task.async do...` between `Async` and `Async::HTTP::Server.new` in the code above.

ioquatix:
Yes, it looks okay to me.

bblimke:
@bruno- @ioquatix thank you both for investigating this and for the fix.

bblimke:
@bruno- @ioquatix actually the build is now failing: https://travis-ci.org/github/bblimke/webmock/jobs/725160832 
`ArgumentError: unknown keyword: transient`

bruno-:
I'm checking this.
My guess is `webmock` is using an older version of `async` or `async-http` gems that don't support that keyword.

bruno-:
@bblimke it seems travis is using `async` `1.24.2` https://travis-ci.org/github/bblimke/webmock/jobs/725160832#L626
That version is from February this year, while `transient` feature was introduced in April.
`async-http` version `>= 0.48.0` specified in webmock gemfile is fine. I don't think that needs to be changed.

Can we bump travis version? Maybe delete travis' cache or something?

bblimke:
@bruno- I now see that it's only a problem on Ruby 2.3 and 2.4 which are likely not compatible with latest async version.

2.3 and 2.4 are EOL so I guess I can drop support for them anyway.

ioquatix:
I have an unpublished policy for open source, I do not support EOL Rubies.

bruno-:
@bblimke looking forward to next webmock gem release that contains these fixes.

bblimke:
This has now been released as version 3.9.0

PR #902, opened 2020-09-08
Fix async http adapter warning

bruno-:
Problem: when using async http adapter rspec outputs the following warning
`lib/webmock/http_lib_adapters/async_http_client_adapter.rb:154: warning: instance variable @alpn_protocol not initialized`
Reason for that is: `@alpn_protocol` is never defined.

Solution: don't use `@alpn_protocol`


bblimke:
Thank you @bruno-

Issue #907, opened 2020-10-12
Global stubs are not thread-safe


adam-harwood:
We have specs that use VCR (hooking into webmock) to test some complex interactions. We stub a significant number (~1000) of requests per test, and the code under test is multi-threaded and processes work asynchronously and in parallel. We also run using JRuby.

In our most complex tests, we see a very rare failure mode where webmock returns a nil response for a request that should be stubbed, and then fails with the error:
```
[2020-10-07T15:45:07.492Z] [2020-10-07T15:45:07.460+00:00][709][2258][connectors][ERROR]: Exception: NoMethodError: undefined method `each_key' for nil:NilClass
[2020-10-07T15:45:07.492Z] /ci/ci_state/ruby_bundle/jruby/2.5.0/gems/webmock-3.8.1/lib/webmock/util/hash_validator.rb:10:in `validate_keys': undefined method `each_key' for nil:NilClass (NoMethodError)
```

After a lot of `puts` statements, I was able to track down the issue to this hash: https://github.com/bblimke/webmock/blob/master/lib/webmock/stub_registry.rb#L25. The global stubs that VCR registers use a hash that is not thread safe, and (in our case) shared between the VCR thread that is stubbing the request and the threads that are actually doing the work and under test. This results in (occasionally) a previously stubbed value being returned instead of the most recent. I can't quite understand the interaction, but somewhere in the VCR/webmock interoperability it seems that a `nil` value is always stubbed first, then the real value is stubbed almost immediately after. Due to the non-thread-safety though, the `nil` value is sometimes returned.

I believe this is rarely seen in the wild because MRI's "threading implementation" hides it. Note that we are using webmock 3.8.1 currently, but I did test with 3.9.1 and the behaviour was the same.

I have tested a fix locally and will open a PR, just wanted to get this issue in here for visibility.

bblimke:
Hi @adam-harwood 

Thank you for the detailed description of the issue and the investigation you have done to trace it.

I look forward to seeing the PR 👍 

> it seems that a `nil` value is always stubbed first

Would you mind clarifying what you mean by `nil` value being stubbed? Do you mean the response is nil?

adam-harwood:
> Would you mind clarifying what you mean by `nil` value being stubbed? Do you mean the response is nil? 

Exactly. Specifically, the yield in register_global_stub is returning nil.

PR #908, opened 2020-10-12
Make global stubs thread-safe

adam-harwood:
This fixes an issue where testing multi-threaded code could sometimes result in incorrect responses being returned. Specifically in the case of VCR, since it seems to mock a nil response first before stubbing the real value, a nil response could sometimes be returned. This was because the hash being used by global stubs is not thread-safe.

I tried to write a test for this, but given the multi-threaded nature, and the fact it doesn't happen on MRI, it was difficult to find anything that failed. The existing tests do at least regression test it though. And I've tested it extensively on the repo where I originally found the issue and it works well.

closes https://github.com/bblimke/webmock/issues/907

### DIFF
--- a/lib/webmock/stub_registry.rb
+++ b/lib/webmock/stub_registry.rb
@@ -23,10 +23,20 @@ module WebMock
       # That way, there's no race condition in case #to_return
       # doesn't run immediately after stub.with.
       responses = {}
+      response_lock = Mutex.new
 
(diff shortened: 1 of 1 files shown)


bblimke:
Thank you @adam-harwood 👍

bblimke:
It's released as 3.9.2

adam-harwood:
> It's released as 3.9.2

Thank you for the quick release!

PR #909, opened 2020-10-16
Make httpclient_adapter threadsafe

adam-harwood:
httpclient_adapter uses some standard Ruby hashes, which produce unexpected results when working in a true multi-threaded (e.g. JRuby) environment. Multiple threads doing puts and deletes on these hashes result in nils sometimes being returned for a properly stubbed request. These errors would manifest as:

`webmock-3.9.2/lib/webmock/http_lib_adapters/httpclient_adapter.rb:113:in build_httpclient_response: undefined method body for nil:NilClass (NoMethodError)`.

Unfortunately these errors are extremely difficult to replicate. I've only been able to reproduce it periodically when running a large suite of VCR tests. Therefore its difficult to write a spec around this. I've run this patch against our local test suite over 10 times and haven't been able to reproduce the error with it, so I'm fairly confident in the fix. Not sure if the synchronization will hurt any other aspects of webmock performance?

This PR is a follow-on from https://github.com/bblimke/webmock/pull/908, and represents the last thread-safety issue we've been able to find with our tests.

### DIFF
--- a/lib/webmock/http_lib_adapters/httpclient_adapter.rb
+++ b/lib/webmock/http_lib_adapters/httpclient_adapter.rb
@@ -43,6 +43,9 @@ if defined?(::HTTPClient)
   end
 
   module WebMockHTTPClients
+
+    REQUEST_RESPONSE_LOCK = Mutex.new
+
     def do_get_block(req, proxy, conn, &block)
       do_get(req, proxy, conn, false, &block)
     end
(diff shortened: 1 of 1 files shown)


bblimke:
Thank you @adam-harwood 

I think this can only hurt other aspects of WebMock performance in a multi-threaded environment, 
which requires that fix anyway, therefore it's better to have this patched, whether it affects performance or not.

bblimke:
It's now released as version 3.9.3

PR #910, opened 2020-10-27
Add: info on adding to gemfile

BKSpurgeon:


bblimke:
@BKSpurgeon Thank you. Perhaps it's better to suggest to add it to `:test` group of the Gemfile.

Most people don't need, and should avoid having, `webmock` included in development or in production.

BKSpurgeon:
I'll squash with a new branch tomorrow.

PR #913, opened 2020-10-27
Update Readme: Add note re: support for the HTTP gem

BKSpurgeon:
Folks when reviewing will be able to quickly see what clients are supported.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,16 +24,19 @@ Features
 Supported HTTP libraries
 ------------------------
 
-* Net::HTTP and libraries based on Net::HTTP (i.e RightHttpConnection, REST Client, HTTParty)
-* HTTPClient
-* Patron
-* EM-HTTP-Request
-* Curb (currently only Curb::Easy)
-* Typhoeus (currently only Typhoeus::Hydra)
-* Excon
-* HTTP Gem
-* Manticore
 * Async::HTTP::Client
+* Curb (currently only Curb::Easy)
+* EM-HTTP-Request
+* Excon
+* HTTPClient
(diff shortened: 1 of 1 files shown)


bblimke:
@BKSpurgeon HTTP is already listed 3 lines above.

BKSpurgeon:
> @BKSpurgeon HTTP is already listed 3 lines above.

You are right! 

I completely missed it! I figured that the real problem is that it is not listed alphabetically; and luckily for me through fortuitous accident, I could in fact use it!

I have updated.

Thank you for reviewing.

README.md
```diff
@@ -24,16 +24,19 @@ Features
...
-* Patron
-* EM-HTTP-Request
+* Async::HTTP::Client
 * Curb (currently only Curb::Easy)
-* Typhoeus (currently only Typhoeus::Hydra)
+* EM-HTTP-Request
 * Excon
-* HTTP Gem
+* HTTPClient
+* [HTTP (The Gem! also known as http.rb)](https://github.com/httprb/http)
```
bblimke, line 32: @BKSpurgeon not sure `The Gem!` is necessary?

README.md
```diff
@@ -24,16 +24,19 @@ Features
...
 * Manticore
-* Async::HTTP::Client
+* Net::HTTP and libraries based on Net::HTTP:
```
bblimke, line 34: @BKSpurgeon WebMock supports these and other clients based on Net::HTTP. This change suggests it only supports the listed 3 libs. I suggest adding `e.g.` or `like`

bblimke:
@BKSpurgeon awesome, thank you!

PR #918, opened 2020-11-05
Fix: Clarify exception message

BKSpurgeon:
# Why this PR?

Right now, users might be confused as to why a hash cannot be accepted
as a response body? See Issue #449 on Github as well as #427.

This seems to have caused considerable consternation
to API users - with some spending many frustrated hours searching for a
solution. Perhaps it can be solved with a more helpful exception
message? This PR hopes to remedy this apparent shortcoming. Perhaps
I shall remedy it with an update to the documentation as well.

I submit it with the utmost respect to the maintainers and community.

Thank you for your efforts, and for reviewing.

Ben

### DIFF
--- a/lib/webmock/response.rb
+++ b/lib/webmock/response.rb
@@ -91,10 +91,10 @@ module WebMock
 
     def ==(other)
       self.body == other.body &&
-      self.headers === other.headers &&
-      self.status == other.status &&
-      self.exception == other.exception &&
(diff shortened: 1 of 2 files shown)


bblimke:
@BKSpurgeon Thank you! This is a very useful change. I believe many developers will appreciate the new error message.

bblimke:
It's now released as version 3.9.4

Issue #915, opened 2020-11-03
Overwriting Test::Unit's teardown causes WebMock to not reset


jessebs:
The standard way of having 'end-of-test' logic is to overwrite the teardown method in the test class. 

Doing so, however, removes the alias that webmock/test_unit creates, causing WebMock.reset! to not be called.

Calling super in the teardown method avoids this behavior, but that is a non-standard approach and also isn't documented in WebMock (that I could find at least)

bblimke:
@jessebs do you have any suggestion how to handle it? I don't think test unit has support for callbacks in the same way as e.g RSpec does.

jessebs:
See PR #919

PR #919, opened 2020-11-06
#915 - Prevent overwriting teardown

jessebs:


bblimke:
Hi @jessebs 

Could you please explain how that works?

jessebs:
Sure.  Take a look at https://test-unit.github.io/test-unit/en/Test/Unit/TestCase.html#teardown-instance_method

Based on that, there are several options for implementing a teardown.  Implementing this way means that a subclass overwriting the teardown won't overwrite the webmock teardown functionality.

I believe that users should be able to overwrite the teardown_with_webmock method to alter that behavior should they really desire it.

bblimke:
Oh wow. Now I'm curious how that works. I mean, how is that `teardown` invocation before method definition implemented internally in test::unit

jessebs:
I just confirmed that overwriting the teardown_with_webmock method in the test class will replace the behavior.

jessebs:
I haven't dug into test::unit to see how it's implemented, but it was definitely a surprise when I found that.

jessebs:
`teardown` is a class method that accepts method names and blocks.  Nothing really fancy but it's a pretty neat way of doing things

bblimke:
@jessebs but there are neither args nor block passed to the `teardown` method in your implementations.
The `teardown_with_webmock` method is declared in the next line, therefore how is it associated with the `teardown` invocation in the previous line?

jessebs:
```
def my_method(*method_names, &callback)
  p method_names
  p callback
end

my_method def example
end
```

jessebs:
Good point.  If it was on the same line that would make sense, but it's not.

Take a look at fixture.rb in the test unit codebase.  I think that holds the answer, but I'm not really able to give it my full attention right now.

jessebs:
Any idea why travis hasn't started building?

bblimke:
Screenshot 2020-11-06 at 18 49 23

jessebs:
@bblimke Travis still appears to be sitting on the build.  Any thoughts?  The PR is no longer a draft

bblimke:
It's released as version 3.9.5. Thank you @jessebs

jessebs:
Thanks for the quick merge!

Issue #171, opened 2012-03-30
Global stubs are given precedence over request stubs


jpignata:
Concerning the change made in https://github.com/bblimke/webmock/commit/a950ab839d3a83a21d284a18858fe1390a5d20d2 where the global stubs are given precedence over any stubs made in the test example:

We were relying on the previous behavior in order to add a catch-all stub that mops up any HTTP requests made and not specifically under test. It feels like this is the more common use case for a global stub rather than using it as a filter to be ran before any stubs are matched. Is there any interest in a pull request that adds a hook to run a block prior to stub matching for the kind of setup that VCR does referenced in the commit message?

PR #920, opened 2020-11-11
Option to invoke local stub before global stub

mkasztelnik:
We have a similar problem as described in #171 in the project currently I'm working on. Since I did not found PR for this issue I created this one. 

By default, nothing is changed in the API but it is possible to give the `after_stubs` option while registering global stub. Then the global stub will be used only when no "local" stub is found.

### DIFF
--- a/spec/acceptance/shared/stubbing_requests.rb
+++ b/spec/acceptance/shared/stubbing_requests.rb
@@ -593,6 +593,23 @@ shared_examples_for "stubbing requests" do |*adapter_info|
           end
         end
       end
+
+      context "when global stub should be invoked last" do
+        before do
+          WebMock.globally_stub_request(:after_local_stubs) do
+            { body: "global stub body" }
+          end
+        end
+
+        it "uses global stub when non-global stub is not defined" do
+          expect(http_request(:get, "http://www.example.com/").body).to eq("global stub body")
+        end
+
+        it "uses non-global stub first" do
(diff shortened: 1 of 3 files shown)


spec/acceptance/shared/stubbing_requests.rb
```diff
@@ -593,6 +593,23 @@ def stub_non_globally
...
       end
+
+      context "when global stub should be invoked last" do
+        before do
+          WebMock.globally_stub_request(:after_stubs) do
+            { body: "global stub body" }
+          end
+        end
+
+        it "uses global stub when non global stub is defined" do
```
bblimke, line 604: I guess that should be `when non-global stub is not defined`?

bblimke:
@mkasztelnik thank you. wow, a PR after 8 years.

`after_stubs` sounds a bit ambiguous to me. Perhaps `before_local_stubs` and `after_local_stubs` would be better?

spec/acceptance/shared/stubbing_requests.rb
mkasztelnik: Yeap - fixed.

mkasztelnik:
I was sure that there is a better name for the `order` values :) Thank you for the great library.

bblimke:
Awesome, thank you!

bblimke:
@mkasztelnik version 3.10 is now released

mkasztelnik:
:heart: :green_heart: :yellow_heart: :blue_heart:

Issue #926, opened 2021-01-28
Webmock breaks Manticore when streaming mode is used


kovyrin:
Manticore allows us to consume response bodies in a [streaming mode](https://github.com/cheald/manticore/blob/master/lib/manticore/response.rb#L104-L112). In that mode, they invoke a given block with small chunks from the response and do not buffer the response body anywhere. Any attempts at reading the response body afterwards result in  a `Manticore::StreamClosedException`. Unfortunately, WebMock's `WebMockManticoreClient#generate_webmock_response` method wants to read the body (for legitimate reasons) and it does not handle the streaming situation well, resulting in any streaming mode requests blowing up with the unhandled `Manticore::StreamClosedException`  error.

# How to reproduce it

First, here is a working case without WebMock: https://gist.github.com/kovyrin/b5aa828694d3b500caeab3afe911c375#file-streaming-rb. This script will execute a request as expected, printing out chunks of content as they are downloaded.

Now, let's add WebMock into the mix: https://gist.github.com/kovyrin/b5aa828694d3b500caeab3afe911c375#file-streaming-with-webmock-rb. This will print out the content chunks as expected, but then, after the request is over, it will blow up from within WebMock.

PR #927, opened 2021-01-28
Fix manticore streaming mode when WebMock is enabled

kovyrin:
This fixes https://github.com/bblimke/webmock/issues/926


bblimke:
Thank you @kovyrin

PR #930, opened 2021-02-24
Only mark it as a real request if it *was* a real request.

ioquatix:
We had an issue reported to async-http about VCR adding requests over and over again: https://github.com/socketry/protocol-http1/issues/11

Here is the provided POC: https://github.com/trevorturk/async-http-vcr

This PR fixes the problem.


bblimke:
@ioquatix thank you for the fix. I hope you don't mind me if I close this PR and merge  https://github.com/bblimke/webmock/pull/931 with the specs. I appreciate your contribution to both PR's :)

ioquatix:
Yes no problem I actually suggested that we make another PR with specs so was expecting that :)

PR #931, opened 2021-02-24
Async http adapter: only consider real requests to be real

tonywok:
This was causing an issue in VCR, as callbacks were always firing because net connect was enabled. In this case, continuously appending recorded cassettes.

A stubbed request is not a real request, and therefore shouldn’t fire callbacks that are only meant for real requests.

Note this was made in collaboration with the fix identified by @ioquatix in #930


bblimke:
Thank you @tonywok and @ioquatix !

bblimke:
It's now released as 3.11.3

PR #932, opened 2021-03-04
Add httpx as supported http client

HoneyryderChuck:
Adding httpx as one of the http clients supporting webmock, with a link for the setup, given that the adapter is self-owned.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ Supported HTTP libraries
   * RightHttpConnection
 * Patron
 * Typhoeus (currently only Typhoeus::Hydra)
+* [httpx](https://honeyryderchuck.gitlab.io/httpx/wiki/Webmock-Adapter) 
 
 Supported Ruby Interpreters
 ---------------------------


bblimke:
Thank you @HoneyryderChuck 

I haven't been aware of httpx gem before, but it looks interesting. 
It's good to know about it in case I need any of its features in the future. 
I appreciate your effort to build support for WebMock in httpx :)

The list of supported http clients listed on WebMock README is for officially supported http clients,
which are guaranteed to be supported by WebMock and this is verified by the adapters passing 
WebMocks acceptance test suite for adapters. 

I have checked Gemfile dependencies and reverse dependencies on rubygems and I believe
httpx is still on a way to get popularity and when it's popular, it will make sense to add an official httpx 
adapter to WebMock and make it as one of the officially supported http clients.

For now I suggest adding information about WebMock on httpx README and it's not there,
therefore even httpx users might not know about the fact that it supports it.

HoneyryderChuck:
Hi @bblimke thank you for taking the time to have a look.

I understand the reasoning, and I don't want to cause more burden to an external project that what you're willing to take it. Just a few considerations:

* webmock isn't mentioned in the Readme, but it [is prominent enough in the project wiki](https://honeyryderchuck.gitlab.io/httpx/wiki/home.html#integrations)
* I followed the model from [faraday, which lists both supported and external adapters](https://lostisland.github.io/faraday/adapters/) Would it make more sense to have a separate section for 3rd party integrations?


It might not be popular compared with other alternatives (it came later), but I have gotten requests in the past about integrating gems for testing/metrics, [including webmock](https://gitlab.com/honeyryderchuck/httpx/-/issues/65), which creates an adoption "chicken or egg" problem for new gems: people want to try your gem, but expect all the other tooling around it to seamlessly work, and not all of them bother to ask. In that regard, I think that a simple "external support" listing could go a long way in breaking out of the popularity monopoly.

However, if you think this is going to increase your support burden, feel free to close it.

bblimke:
@HoneyryderChuck good points. I agree it's difficult to get noticed within many http clients.  
I have added httpx to the README (btw. I followed your PR with lowercase name, but perhaps it's HTTPX?)

Monopoly is probably not the word I'd use :) I think people will eventually switch and start using
a gem that makes a difference and makes their developer lives easier, and if it doesn't, they won't.
Perhaps it's worth updating the project homepage with a big statement to show that difference right away?
I was able to find the features httpx offers, but not from the homepage.

HoneyryderChuck:
@bblimke thx. Lowercase is also fine.

I'll follow up on your suggestion, but feel free to suggest away smth (I was never a great marketer I guess).

I'll close the PR then

PR #934, opened 2021-03-05
Fix copy-paste of JSON instead of XML in spec file

valscion:
Similar to cb3d92965ff422845cecd68f461d807d0b87a0dc ☺️


bblimke:
Thank you @valscion ! :)

PR #936, opened 2021-03-25
Avoid redundantly calling em-http-request middleware

### LINKED ISSUE #899: em-http-request middlewares get double-called
Specifically, the `#request` methods are invoked in two places.

1. em-http-request calls them from `EventMachine::HttpClient#connection_completed`: https://github.com/igrigorik/em-http-request/blob/157d5ff281c503656192825c388b28e7f35e04ce/lib/em-http/client.rb#L57-L59
2. webmock calls them from `EventMachine::WebMockHttpClient#build_request_signature`: https://github.com/bblimke/webmock/blob/382d84ca586fc87755d5657616d7a492421ee11b/lib/webmock/http_lib_adapters/em_http_request_adapter.rb#L169-L171

The request signature is built at the start of `EventMachine::WebMockHttpConnection#activate_connection` here:
https://github.com/bblimke/webmock/blob/382d84ca586fc87755d5657616d7a492421ee11b/lib/webmock/http_lib_adapters/em_http_request_adapter.rb#L56

This happens prior to duplicating the logic of `EventMachine::HttpConnection#activate_connection` ([ref](https://github.com/igrigorik/em-http-request/blob/157d5ff281c503656192825c388b28e7f35e04ce/lib/em-http/http_connection.rb#L120-L130)) to call `#finalize_request` here:
https://github.com/bblimke/webmock/blob/382d84ca586fc87755d5657616d7a492421ee11b/lib/webmock/http_lib_adapters/em_http_request_adapter.rb#L60-L69

Then the `#finalize_request` call  hits the superclass `EventMachine::HttpConnection#finalize_request`, which adds a callback to `EventMachine::HttpClient#connection_completed`: https://github.com/igrigorik/em-http-request/blob/157d5ff281c503656192825c388b28e7f35e04ce/lib/em-http/http_connection.rb#L159

Which, as noted at the start, runs the `#request` middleware again.

The middleware handling was added waaay back in #111. Since the `#request` hook can amend the request headers & body, it makes sense that you'd want it to run before building & checking the webmock signature. But the circuitous control flow means that if the middleware has a side-effect (e.g., adding another callback), those side-effects will get triggered twice, which can lead to some really bizarre (and incorrect) behavior in the test suite.

ajvondrak:
This appears to fix #899. em-http-request is such a ball of spaghetti
that I have my doubts, but the tests are green.

The patch is based on the observation that the em-http-request call
sequence goes:

1. `HttpConnection#activate_connection`
2. `HttpClient#connection_completed`
3. `HttpClient#send_request`

The `#connection_completed` method is normally responsible for calling
the request middleware before passing the adjusted headers/body to
`#send_request` (cf.
https://github.com/igrigorik/em-http-request/blob/34919d760b940dfd27159bbebce5c4dac9845eb6/lib/em-http/client.rb#L53-L62).
But webmock needs the request signature all the way back in
`#activate_connection`, which we want to compute using the
middleware-adjusted headers/body.

So instead, we let `#activate_connection` handle building the request
signature, whose result will be memoized. Then when we monkey patch
`#connection_completed` to look up the headers/body from that signature
without recomputing the values.

This required some tweaking to `#build_request_signature`, where it was
actually failing to follow the same algorithm as em-http-request.
Basically it just means calling em-http-request's internal
`#build_request` method + making sure not to clobber the `headers` local
variable (which was probably not intentional in the first place).

### DIFF
--- a/lib/webmock/http_lib_adapters/em_http_request_adapter.rb
+++ b/lib/webmock/http_lib_adapters/em_http_request_adapter.rb
@@ -99,6 +99,11 @@ if defined?(EventMachine::HttpClient)
         end
       end
 
+      def connection_completed
+        @state = :response_header
+        send_request(request_signature.headers, request_signature.body)
+      end
+
       def send_request(head, body)
         WebMock::RequestRegistry.instance.requested_signatures.put(request_signature)
 
@@ -164,7 +169,7 @@ if defined?(EventMachine::HttpClient)
       end
 
       def build_request_signature
-        headers, body = @req.headers, @req.body
+        headers, body = build_request, @req.body
 
         @conn.middleware.select {|m| m.respond_to?(:request) }.each do |m|
           headers, body = m.request(self, headers, body)
@@ -177,8 +182,6 @@ if defined?(EventMachine::HttpClient)
         uri.query = encode_query(@req.uri, query).slice(/\?(.*)/, 1)
 
         body = form_encode_body(body) if body.is_a?(Hash)
-
-        headers = @req.headers
 
         if headers['authorization'] && headers['authorization'].is_a?(Array)
           headers['Authorization'] = WebMock::Util::Headers.basic_auth_header(headers.delete('authorization'))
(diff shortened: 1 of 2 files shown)


bblimke:
Thank you @ajvondrak

After checking the code I can see it was tricky to figure out the right approach.

bblimke:
It's now released as 3.11.2

PR #947, opened 2021-07-13
fix async adapter on Windows

ahorek:
even if Socket::AF_UNIX is defined on Windows, it cannot be used, so let's use Socket::AF_INET as a fallback


bblimke:
@ahorek sorry, I missed it.

ahorek:
np, thanks!

PR #951, opened 2021-08-18
Change `blacklist` to `denylist` in documentation

clinejj:
This updates documentation to use the more inclusive `denylist` verbiage instead of `blacklist`.


bblimke:
Thank you @clinejj

PR #952, opened 2021-08-20
Add RequestStub#to_return_json

tjefferson08:
Make stubbing out json responses a little bit more convenient by

1) JSON-ifying Hash bodies
2) applying the Content-Type: application/json header

Some relevant discussion in #449

### DIFF
--- a/lib/webmock/request_stub.rb
+++ b/lib/webmock/request_stub.rb
@@ -24,6 +24,21 @@ module WebMock
     end
     alias_method :and_return, :to_return
 
+    def to_return_json(*response_hashes)
+      raise ArgumentError, '#to_return_json does not support passing a block' if block_given?
+
+      json_response_hashes = [*response_hashes].flatten.map do |resp_h|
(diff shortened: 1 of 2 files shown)


JackDanger:
Makes sense. This is how I generally use webmock

bblimke:
@tjefferson08 thank you. that looks useful! 

Perhaps it should allow overriding content_type in case people use `application/vnd.api+json` or similar?
E.g. `{content_type: 'application/json'}.merge headers.to_h` ?

tjefferson08:
@bblimke thanks for having a look! 

I think letting folks override `content_type` makes perfect sense! 

Although I suppose it does eat away at the positioning as a convenience helper. (at that point we're solely calling`.to_json` for you)

In any case, I'll send another commit; IMO it's more correct the way you've described.

ChrisBr:
@bblimke this is super useful 👍 Any plan to release a new version of webmock?

PR #960, opened 2021-12-10
Start connection in `Net::HTTP` adapter for mocked requests

ojab:
Instrumentation like Sentry only instruments requests when
`Net::HTTP.started?` because `#request` is called twice if connection
is not started yet. This change allows this instrumentation to work
with mocked requests.

See https://github.com/ruby/net-http/blob/master/lib/net/http.rb#L1529-L1534

### DIFF
--- a/lib/webmock/http_lib_adapters/net_http.rb
+++ b/lib/webmock/http_lib_adapters/net_http.rb
@@ -77,6 +77,8 @@ module WebMock
           WebMock::RequestRegistry.instance.requested_signatures.put(request_signature)
 
           if webmock_response = WebMock::StubRegistry.instance.response_for_request(request_signature)
+            return start_without_connect { request(request, body, &block) } unless started?
+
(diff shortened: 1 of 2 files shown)


ojab:
Ready to review because CI waits for approval, `rspec spec/acceptance/net_http/` is green locally.

bblimke:
@ojab I'm sorry for the delay.

Is that related specifically to Sentry which checks for `Net::HTTP.started?` . My concern is that in all other cases it will unnecessarily call request twice without any specific need.

ojab:
Sentry prompted me to look into that and I didn't saw any issues about it in other places, but overall it should make mocking more correct and shouldn't bring much overhead.

rzane:
#992 also fixes this, if I'm not mistaken.

bblimke:
@ojab is this PR still required since https://github.com/bblimke/webmock/pull/992 has already been released?

Issue #863, opened 2019-11-27
`WebMock::Util::Headers.normalize_headers` downcases characters in header keys


ulyssesrex:
This stub, chained with `with()`, will never match its intended request...

```ruby
stub_request(:post, /urlmatcher/).
  with(headers: { 
    "SOAPAction" => "https://some.soap.endpoint/where/things/happen" 
  })
```

because the stub's `HeadersPattern` has been normalized and no longer matches the given header. Note `Soapaction` instead of `SOAPAction`.
```
#<WebMock::HeadersPattern:0x00007fe38db8b078                                                                                                             
     @pattern={"Soapaction"=>"https://www.assuresign.net/Services/DocumentNOW/Envelopes/IEnvelopeService/CreateEnvelope"}>
```

bblimke:
@ulyssesrex thank you for pointing that. SOAPAction is indeed a culprit. I guess this can be changed in WebMock but with major version change in WebMock as I it would break many existing WebMock setups, where people set header expectations using e.g lowercase.

ioquatix:
The normalized form of headers which is enforced by HTTP/2 is lowercase.

bblimke:
Thank you for the input @ioquatix It's only for HTTP/2 right?

ioquatix:
It's also for HTTP/1.

https://stackoverflow.com/questions/5258977/are-http-headers-case-sensitive

HTTP/2 just makes it an error and servers will drop the connection.

ioquatix:
Just FYI, this will become standard rack behaviour too: https://github.com/rack/rack/issues/1592

PR #964, opened 2022-01-05
Keep underscores when normalizing headers

### LINKED ISSUE #474: `WebMock::Util::Headers.normalize_headers` replaces underscores with dashes
WebMock is unable to correctly mock server responses that include underscores in the header names. Is there a specific reason for this behaviour? I checked the HTTP specification and underscores are not an illegal character.

The offending line is here:

https://github.com/bblimke/webmock/blob/6b5374c50b288c7086a3eecd098f78507d6f8994/lib/webmock/util/headers.rb#L10

mvz:
Some servers distinguish underscores from dashes in request headers. To allow users to test that the correct character is used, checking the sent headers should make the same distinction.

With this change, webmock no longer normalizes underscores to dashes for headers defined using strings.

Since symbols are commonly used with underscores to specify headers that use dashes, headers specified with symbol keys are still normalized to dashes.

Fixes #474.

### DIFF
--- a/lib/webmock/util/headers.rb
+++ b/lib/webmock/util/headers.rb
@@ -9,7 +9,8 @@ module WebMock
       def self.normalize_headers(headers)
         return nil unless headers
         array = headers.map { |name, value|
-          [name.to_s.split(/_|-/).map { |segment| segment.capitalize }.join("-"),
+          name = name.to_s.tr("_", "-") if name.is_a? Symbol
+          [name.to_s.split(/(_|-)/).map { |segment| segment.capitalize }.join,
            case value
             when Regexp then value
             when Array then (value.size == 1) ? value.first.to_s : value.map {|v| v.to_s}.sort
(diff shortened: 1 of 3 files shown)


bblimke:
@mvz thank you for the PR. I'm not sure about the new logic though.

mvz:
>  It has a different behaviour for symbols than strings based on assumption "commonly used with underscores" yet strings are not.

Yes, that is on purpose since symbols with dashes are harder to write so many developers will probably have symbols used like that. However, I'm all for being more strict here and just unifying the behavior.

In fact, given https://github.com/bblimke/webmock/issues/863#issuecomment-571339772, normalization could then become just a simple `.downcase` on the stringified key.

> Change like that will require a major version bump anyway.

To make this smoother, this could also be introduced as an optional change controlled by a setting. The old behavior can than be removed with a major version bump later.

PR #965, opened 2022-01-13
Make CI green again

ojab:
httpstat.us switched to HTTPS and returning 301 for plain HTTP and now
there is no `X-Powered-By` header, use HTTPS + `Server` header instead.

`http.rb` >= 5.0.2 expects stream result to be in the proper encoding after https://github.com/httprb/http/pull/653, force encoding in `Streamer`.

### DIFF
--- a/lib/webmock/http_lib_adapters/http_rb/response.rb
+++ b/lib/webmock/http_lib_adapters/http_rb/response.rb
@@ -24,7 +24,10 @@ module HTTP
         elsif HTTP::VERSION < "3.0.0"
           Body.new(Streamer.new(webmock_response.body), webmock_response.body.encoding)
         else
-          Body.new(Streamer.new(webmock_response.body), encoding: webmock_response.body.encoding)
+          Body.new(
+            Streamer.new(webmock_response.body, encoding: webmock_response.body.encoding),
+            encoding: webmock_response.body.encoding
+          )
         end
 
         return new(status, "1.1", headers, body, uri) if HTTP::VERSION < "1.0.0"
--- a/lib/webmock/http_lib_adapters/http_rb/streamer.rb
+++ b/lib/webmock/http_lib_adapters/http_rb/streamer.rb
@@ -1,8 +1,9 @@
 module HTTP
   class Response
     class Streamer
-      def initialize(str)
+      def initialize(str, encoding: Encoding::BINARY)
         @io = StringIO.new str
+        @encoding = encoding
       end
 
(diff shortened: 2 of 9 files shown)


ojab:
Ok, apparently httpstat.us made some changes a month ago and now there is no `Content-Length` header. 
httpbin.org doesn't return status message, only status code.

I guess we can run https://github.com/aaronpowell/httpstatus/pkgs/container/httpstatus in GHA, but it would complicate local development. 

Any hints what's the desired solution? I can try to find another service that could be used or file a bug for https://github.com/aaronpowell/httpstatus, but it's fragile anyway.

ojab:
Filed an issue https://github.com/aaronpowell/httpstatus/issues/105

bblimke:
@ojab I wonder why the builds fail on Ruby 3.0.

ojab:
`spec/acceptance/async_http_client/async_http_client_spec.rb` hangs :crying_cat_face: 
I'm gonna take a look at it this week.

ojab:
Okay, looks like Async is incompatible with RSpec in ruby >= 3.0.
```ruby
require 'async'
require 'async/http'

RSpec.describe 'qwe' do
  let(:endpoint) { Async::HTTP::Endpoint.parse('http://www.example.com') }
  let(:protocol) { Async::HTTP::Protocol::HTTP10 }
  let(:requests_count) { 1 }

  subject do
    puts "In subject"
    responses = {}
    Async do |task|
      puts "In Async"
      Async::HTTP::Client.open(endpoint, protocol) do |client|
        puts "In Client"
        requests_count.times do |index|
          response = client.get "/foo#{index}"
          responses[index] = response_to_hash(response)
        end
      end
    end
    responses
  end

  specify do
    puts "Running!"
    expect(subject).not_to be_nil
  end
end
```
this spec in empty project with Gemfile
```ruby
# frozen_string_literal: true

source "https://rubygems.org"

# gem "rails"

gem 'async-http'
gem 'rspec'
```
hangs the same way. async-2.0 switched to fiber scheduler and dropped ruby<3.0 support, so I don't see value in reporting it.

Anyway, not a webmock bug, workarounded by replacing `let` with `let!` in the last commit.

ojab:
And some RSpec changes broken two specs, fixed in the last commit.
CI in my fork is green https://github.com/ojab/webmock/actions/runs/1824337706 with all the changes.

bblimke:
Thank you @ojab

Issue #946, opened 2021-07-07
Curb mocks fail with headers that include more than one colon ":"


mrbuzz:
When using headers that contain more than one ":" character the resulting stubbed request will only contain a portion of the original header.  
``` ruby
RSpec.describe "#webmock_test" do
  before do
    stub_request(:get, "https://www.example.com/?a=b&c=d").
    with(
      headers: {
     'Content-Length'=>'0',
     'Referer'=>'https://www.example.com?a=b&c=d'
      }).
    to_return(status: 200, body: "", headers: {})
  end

  it "send a request with referer header" do
    client = Curl::Easy.new.tap do |http_client|
      http_client.headers = ["Referer: https://www.example.com?a=b&c=d", "Content-Length: 0"]
      http_client.url = "https://www.example.com?a=b&c=d"
    end
    client.http("GET")
  end
end
```

Output:

```
Failures:

  1) #webmock_test send a request with referer header
     Failure/Error: client.http("GET")
     
     WebMock::NetConnectNotAllowedError:
       Real HTTP connections are disabled. Unregistered request: GET https://www.example.com/?a=b&c=d with headers {'Content-Length'=>'0', 'Referer'=>'https'}
     
       You can stub this request with the following snippet:
     
       stub_request(:get, "https://www.example.com/?a=b&c=d").
         with(
           headers: {
          'Content-Length'=>'0',
          'Referer'=>'https'
           }).
         to_return(status: 200, body: "", headers: {})
     
       registered request stubs:
     
       stub_request(:get, "https://www.example.com/?a=b&c=d").
         with(
           headers: {
          'Content-Length'=>'0',
          'Referer'=>'https://www.example.com?a=b&c=d'
           })
     
       ============================================================
```
I think this can be fixed by changing line 131 in the method referenced below 
https://github.com/bblimke/webmock/blob/a3e1c2c6f1257634c05a38c6832fd0e4c0bae853/lib/webmock/http_lib_adapters/curb_adapter.rb#L128-L138
to 
``` ruby
name, value = header.split(":", 2).map(&:strip) 
``` 
this will prevent the header from being split in more than 2 pieces .

bblimke:
@mrbuzz thank you for reporting that. Yes, it looks like a good solution. Would you fancy submitting a PR?

mrbuzz:
@bblimke Yes! I'll submit a PR, hopefully in the next days.

PR #968, opened 2022-02-01
Limit the number of splits when parsing headers in the Curb adapter

mrbuzz:
When using headers that contain more than one ":" character the resulting stubbed request will only contain a portion of the original header. This change will prevent the header from being split in more than 2 pieces.

Fixes: [#946 ](https://github.com/bblimke/webmock/issues/946)

### DIFF
--- a/lib/webmock/http_lib_adapters/curb_adapter.rb
+++ b/lib/webmock/http_lib_adapters/curb_adapter.rb
@@ -128,7 +128,7 @@ if defined?(Curl)
       def headers_as_hash(headers)
         if headers.is_a?(Array)
           headers.inject({}) {|hash, header|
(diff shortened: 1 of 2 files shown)


pboling:
@bblimke thoughts?  Also would this allow us to remove the curb compatibility warning on recent versions of curb?

bblimke:
@mrbuzz Thank you!

nikolasgd:
@mrbuzz 
I am using Curb 1.0.6 with webmock. Since the issues seems to be resolved, can we get rid of the warning message?

nikolasgd:
I have made a pull request for this:

https://github.com/bblimke/webmock/pull/1080

PR #969, opened 2022-02-15
Allow content type headers to specify a charset

Timmitry:
When using the HTTP gem with a command like `HTTP.post(url, json: { a: 1 })`, the `Content-Type`-header is set to `application/json; charset=UTF-8`. This currently prevents me from using the `RequestBodyDiff`-method, as a check for `json_headers?` is made, which returns false if the content type isn't exactly `application/json`.

Specifying the charset in the content type [should be allowed](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type#syntax). This PR therefore suggests just to check whether the content type starts with `application/json` instead of requiring an exact match.

The same is done for the `url_encoded?` helper method, checking for the `application/x-www-form-urlencoded` content type.

### DIFF
--- a/lib/webmock/request_signature.rb
+++ b/lib/webmock/request_signature.rb
@@ -35,11 +35,11 @@ module WebMock
     alias == eql?
 
     def url_encoded?
-      !!(headers && headers['Content-Type'] == 'application/x-www-form-urlencoded')
+      !!(headers&.fetch('Content-Type', nil)&.start_with?('application/x-www-form-urlencoded'))
     end
 
     def json_headers?
-      !!(headers && headers['Content-Type'] == 'application/json')
+      !!(headers&.fetch('Content-Type', nil)&.start_with?('application/json'))
     end
 
     private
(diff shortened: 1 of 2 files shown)


lib/webmock/request_signature.rb
```diff
@@ -35,11 +35,11 @@ def eql?(other)
     alias == eql?
 
     def url_encoded?
-      !!(headers && headers['Content-Type'] == 'application/x-www-form-urlencoded')
+      !!(headers && headers['Content-Type'].start_with?('application/x-www-form-urlencoded'))
```
dklotz, line 38: What happens if there is no Content-Type header? Maybe it would be better to make it nil-safe like this: 
```suggestion
      !!(headers && headers['Content-Type']&.start_with?('application/x-www-form-urlencoded'))
```

cgrothaus:
I raised this as issue #971 to increase visibility.

lib/webmock/request_signature.rb
Timmitry: You're absolutely right, I missed this case 👍 

I hope the usage of safe navigation is okay (it isn't used anywhere in the project yet) - if not, I can change it to use more `&&` chaining 😉

bblimke:
Thank you @Timmitry !

Issue #955, opened 2021-09-15
net_http_connect_on_start: true breaks disable_net_connect! and stub_request?


dmolesUC:
## Summary

On the recommendation of the [Capybara docs](https://github.com/teamcapybara/capybara#gotchas), we've been setting `net_http_connect_on_start: true` in order to avoid having WebMock run us out of file handles during Selenium tests.

We've recently discovered, however, that this has the effect of allowing WebMock to work only if there's an actual, reachable service running on the specified host and port -- for a nonexistent hostname, or an unreachable host/port combination, net connections are not disabled and requests are not stubbed.

Per [this comment on #914](https://github.com/bblimke/webmock/issues/914#issuecomment-810134233) something like this is apparently expected behavior, but the [README](https://github.com/bblimke/webmock#connecting-on-nethttpstart) isn't very explicit. Maybe it's implied if you know a lot about `Net::HTTP` internals?

But I don't really care whether I have long-running connections, I just want to intercept requests, and not to run out of file handles. Is there any way around this?

## Steps to reproduce

Given the following test:

```ruby
require 'webmock/rspec'

describe :disable_net_connect! do
  shared_examples 'correctly' do |test_uri|
    it "intercepts Net::HTTP connections" do
      expect { Net::HTTP.get(test_uri) }.to raise_error(WebMock::NetConnectNotAllowedError)
    end

    it "can stub Net::HTTP requests" do
      http_status = 403
      stub_request(:get, test_uri).to_return(status: http_status)

      response = Net::HTTP.get_response(test_uri)
      expect(response.code).to eq(http_status.to_s)
    end
  end

  describe 'without net_http_connect_on_start' do
    around(:each) do |example|
      WebMock.disable_net_connect!
      example.run
      WebMock.allow_net_connect!
    end

    describe 'with real URI' do
      it_behaves_like 'correctly', URI.parse('http://example.org/')
    end

    describe 'with nonexistent URI' do
      it_behaves_like 'correctly', URI.parse('http://nonexistent-hostname:9999/')
    end

    describe 'with unreachable URI' do
      # NOTE: `other-local-hostname` is mapped to 127.0.0.1 in /etc/hosts
      it_behaves_like 'correctly', URI.parse('http://other-local-hostname:9999/')
    end
  end

  describe 'with net_http_connect_on_start' do
    around(:each) do |example|
      WebMock.disable_net_connect!(
        net_http_connect_on_start: true
      )
      example.run
      WebMock.allow_net_connect!
[...]

Issue #914, opened 2020-10-29
net_http_connect_on_start: true does not correctly stub out URLs with subdomains


kryptonat:
I'm running into [file descriptor leakages](https://github.com/teamcapybara/capybara/issues/2322) and found out that you can set `net_http_connect_on_start` option to `true` as recommended [here](https://github.com/terrchen/gitlab/commit/a8abf6da316a2d151cfb9e70ff1adb9a9986ddfd).

However when this option is enabled, it's not able to correctly stub out a request with subdomains. This seems like a bug.

**Actual** 
```ruby
    WebMock.disable_net_connect!(net_http_connect_on_start: true)

    # Stubbing URL with subdomain, results in an error.
    stub_request(:get, 'http://api.foobar.com').to_return(body: 'hi')
    Net::HTTP.get(URI('http://internal-api.foobar.com'))
    # => SocketError: Failed to open TCP connection to api.foobar.com:80 (getaddrinfo: nodename nor servname provided, or not known)
 
    # Stubbing URL with no subdomain, can be stubbed successfully.
    stub_request(:get, 'http://foobar.com').to_return(body: 'hi')
    Net::HTTP.get(URI('http://foobar.com'))
    # => "hi"

    # When net_http_connect_on_start defaults to false...
    WebMock.disable_net_connect!

    # Stubbing URL with subdomain, can be stubbed successfully.
    Net::HTTP.get(URI('http://internal-api.foobar.com'))
    # => "hi"
```

**Expected**
```ruby
    WebMock.disable_net_connect!(net_http_connect_on_start: true)
    stub_request(:get, 'http://api.foobar.com').to_return(body: 'hi')
    # => "hi"
```

kryptonat:
Just stepped through the code a bit, looks like enabling `net_http_connect_on_start` invokes `connect` from `Net::HTTP` which consequently creates a `TCPSocket` with the stubbed URL

https://github.com/ruby/net-http/blob/1502552ce444c122b7920d8254c79a6cd2e94155/lib/net/http.rb#L987

If that URL host cannot be resolved, it throws the SocketError `(getaddrinfo: nodename nor servname provided, or not known)`. I'm not sure if it's intentionally written this way but if that's the case, enabling `net_http_connect_on_start` is not an option for resolving the file descriptor leakages if any of the stubbed requests do not resolve in a successful DNS lookup.

privorotskii:
Actual issue for me as well

bblimke:
[...]

PR #975, opened 2022-05-22
Allow specifying when Net::HTTP should connect on start

rzane:
Currently, `net_http_connect_on_start` is all or nothing. Many people only want to connect Selenium's requests on start so that WebMock and Capybara play nicely together.

Using `net_http_connect_on_start: true` has some undesirable behaviors (see #914 and #955). I think that this PR provides a workaround for those issues. For most people, I think `net_http_connect_on_start: %w(localhost 127.0.0.1)` will be a more ideal configuration.

This PR is backwards compatible.


bblimke:
Thank you @rzane 

I have merged it and adjusted to avoid conflicts and be compatible with your other PR related to leaky file descriptors.

PR #976, opened 2022-05-24
Fix leaky file descriptors and reuse socket for persistent connections

rzane:
Take the following example:

```ruby
require 'net/http'
require 'webmock'
WebMock.enable!
WebMock.allow_net_connect!

http = Net::HTTP.new('example.org')
http.start
http.get('/')
http.get('/')
http.get('/')
http.finish
```

This example would create 3 sockets and only close one of them. That's because `start_with_connect_without_finish` spawns a new connection for each request, and never closes the previous one.

This PR also makes it so that when using a persistent connection, the `@socket` will be reused across requests. I believe this should totally fix the common "gotcha" that people encounter when combining WebMock with Capybara _without_ needing to enable `net_http_connect_on_start`.

I'd recommend reviewing this PR commit-by-commit. I tried to provide pretty detailed explanations for the changes as I made them.

### DIFF
--- a/lib/webmock/http_lib_adapters/net_http.rb
+++ b/lib/webmock/http_lib_adapters/net_http.rb
@@ -132,6 +132,10 @@ module WebMock
 
 
         def start_with_connect_without_finish
+          if @socket
+            @socket.close
+            @socket = nil
+          end
           do_start
         end
 
--- a/lib/webmock/http_lib_adapters/net_http.rb
+++ b/lib/webmock/http_lib_adapters/net_http.rb
@@ -101,9 +101,8 @@ module WebMock
               if WebMock::Config.instance.net_http_connect_on_start
                 super_with_after_request.call
               else
-                start_with_connect_without_finish {
-                  super_with_after_request.call
-                }
+                start_with_connect_without_finish
+                super_with_after_request.call
               end
             else
               start_with_connect {
@@ -132,15 +131,8 @@ module WebMock
         end
 
 
-        def start_with_connect_without_finish  # :yield: http
-          if block_given?
-            begin
-              do_start
-              return yield(self)
-            end
-          end
+        def start_with_connect_without_finish
           do_start
-          self
         end
 
         alias_method :start_with_connect, :start
--- a/lib/webmock/http_lib_adapters/net_http.rb
+++ b/lib/webmock/http_lib_adapters/net_http.rb
@@ -98,12 +98,8 @@ module WebMock
               after_request.call(response)
             }
             if started?
-              if WebMock::Config.instance.net_http_connect_on_start
-                super_with_after_request.call
-              else
-                start_with_connect_without_finish
-                super_with_after_request.call
-              end
+              ensure_actual_connection
+              super_with_after_request.call
             else
               start_with_connect {
                 super_with_after_request.call
@@ -131,12 +127,8 @@ module WebMock
         end
 
 
-        def start_with_connect_without_finish
-          if @socket
-            @socket.close
-            @socket = nil
-          end
-          do_start
+        def ensure_actual_connection
+          do_start if @socket.is_a?(StubSocket)
         end
 
         alias_method :start_with_connect, :start
--- a/lib/webmock/http_lib_adapters/net_http.rb
+++ b/lib/webmock/http_lib_adapters/net_http.rb
@@ -119,12 +119,14 @@ module WebMock
           raise IOError, 'HTTP session already opened' if @started
           if block_given?
             begin
+              @socket = Net::HTTP.socket_type.new
               @started = true
               return yield(self)
             ensure
               do_finish
             end
           end
+          @socket = Net::HTTP.socket_type.new
           @started = true
           self
         end
@@ -258,6 +260,7 @@ class StubSocket #:nodoc:
 
   class StubIO
     def setsockopt(*args); end
+    def peer_cert; end
   end
 end
 
--- a/spec/acceptance/net_http/net_http_shared.rb
+++ b/spec/acceptance/net_http/net_http_shared.rb
@@ -26,20 +26,20 @@ shared_examples_for "Net::HTTP" do
 
     it "should connect only once when connected on start", net_connect: true do
       @http = Net::HTTP.new('localhost', port)
-      socket_id_before_request = socket_id_after_request = nil
+      socket_before_request = socket_after_request = nil
       @http.start {|conn|
-        socket_id_before_request = conn.instance_variable_get(:@socket).object_id
+        socket_before_request = conn.instance_variable_get(:@socket)
         conn.request(Net::HTTP::Get.new("/"))
-        socket_id_after_request = conn.instance_variable_get(:@socket).object_id
+        socket_after_request = conn.instance_variable_get(:@socket)
       }
 
-      if !defined?(WebMock::Config) || WebMock::Config.instance.net_http_connect_on_start
-        expect(socket_id_before_request).not_to eq(nil.object_id)
-        expect(socket_id_after_request).not_to eq(nil.object_id)
-        expect(socket_id_after_request).to eq(socket_id_before_request)
+      if WebMock::Config.instance.net_http_connect_on_start
+        expect(socket_before_request).to be_a(Net::BufferedIO)
+        expect(socket_after_request).to be_a(Net::BufferedIO)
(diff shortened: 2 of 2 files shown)


kbruccoleri:
👍 I have also been experiencing this issue, this seemed to do the trick for me.

slayer (approved):

bblimke:
Hi @rzane. Thank you for the PR with very descriptive commit messages. It's way cleaner and understandable now. 👍

jdufresne:
Hi, This change resulted in a new error for my test suite. I tracked it down to commit 18dc8a9ec2b3210a680a0a1c1642da8271d5fa8e.

The error appears as shown below. At first glance it looks like the new stub objects are missing some method.

     NoMethodError:
       undefined method `ssl_version' for #<StubSocket::StubIO:0x00007f08a997b388>

             { version: @socket.io.ssl_version, cipher: @socket.io.cipher[0] }
                                  ^^^^^^^^^^^^
     # ./.bundle/ruby/3.1.0/gems/activemerchant-1.126.0/lib/active_merchant/net_http_ssl_connection.rb:8:in `ssl_connection'
     # ./.bundle/ruby/3.1.0/gems/activemerchant-1.126.0/lib/active_merchant/connection.rb:82:in `block (2 levels) in request'
     # ./.bundle/ruby/3.1.0/gems/activemerchant-1.126.0/lib/active_merchant/connection.rb:80:in `block in request'
     # ./.bundle/ruby/3.1.0/gems/activemerchant-1.126.0/lib/active_merchant/network_connection_retries.rb:24:in `block in retry_exceptions'
     # ./.bundle/ruby/3.1.0/gems/activemerchant-1.126.0/lib/active_merchant/network_connection_retries.rb:52:in `retry_network_exceptions'
     # ./.bundle/ruby/3.1.0/gems/activemerchant-1.126.0/lib/active_merchant/network_connection_retries.rb:23:in `retry_exceptions'
     # ./.bundle/ruby/3.1.0/gems/activemerchant-1.126.0/lib/active_merchant/connection.rb:75:in `request'
     # ./.bundle/ruby/3.1.0/gems/activemerchant-1.126.0/lib/active_merchant/posts_data.rb:74:in `raw_ssl_request'
     # ./.bundle/ruby/3.1.0/gems/activemerchant-1.126.0/lib/active_merchant/posts_data.rb:44:in `ssl_request'
     # ./.bundle/ruby/3.1.0/gems/activemerchant-1.126.0/lib/active_merchant/posts_data.rb:40:in `ssl_post'
     # ./.bundle/ruby/3.1.0/gems/activemerchant-1.126.0/lib/active_merchant/billing/gateways/authorize_net.rb:817:in `commit'
     # ./.bundle/ruby/3.1.0/gems/activemerchant-1.126.0/lib/active_merchant/billing/gateways/authorize_net.rb:120:in `authorize'

lib/webmock/http_lib_adapters/net_http.rb
```diff
@@ -119,26 +114,21 @@ def start_without_connect
           raise IOError, 'HTTP session already opened' if @started
           if block_given?
             begin
+              @socket = Net::HTTP.socket_type.new
```
jdufresne, line 117: FWIW, `Net::HTTP.socket_type` is obsolete:

https://github.com/ruby/net-http/blob/196f3d70f9ce154dc63a86a2cdbf1aec4df4982e/lib/net/http.rb#L575-L577

bblimke:
@jdufresne thank you for reporting. 

I wonder if this is related to Ruby 3.1 or happens in earlier versions as well.

I guess it's not easy for you to provide a code sample or spec to reproduce it, since it's active merchant?

Am I right that it happens when the request is stubbed and not for the real request?

Perhaps StubIO should have more methods implemented, but then why has it not been happening before. Perhaps StubIO was not used.

@rzane any ideas?

jdufresne:
> Am I right that it happens when the request is stubbed and not for the real request?

That is right. This only occurs in the test suite when the request is stubbed.

> Perhaps StubIO should have more methods implemented, but then why has it not been happening before. Perhaps StubIO was not used.

IIUC, StubIO wasn't introduced until 18dc8a9ec2b3210a680a0a1c1642da8271d5fa8e which is why it was not used before. This has only been included in the most recent release. Downgrading fixes the issue for me.

Adding the two methods `ssl_version` and `cipher` fixes the issue for me, but: 

1. Do other users rely on another missing method?
2. Is just adding blank methods the way we want to go here?
3. Does `ssl_version` apply to both HTTP and HTTPS connection or just HTTPS?

bblimke:
StubIO was introduced way ago https://github.com/bblimke/webmock/commit/c4b232badc7ac7a11c1d2ac04dd455b1cd63b2e1 but obviously now it's being set in more scenarios than before and I wonder what are these scenarios, where despite stubbed request, normal socket was used instead of stubbed one with StubIO. I would have thought that StubIO was always used for stubbed requests, but obviously not.

rzane:
The issue @jdufresne is seeing is coming from the fact that ActiveMerchant are getting information from the `@socket`, which is a private instance variable. See [here](https://github.com/activemerchant/active_merchant/blob/3407c02c18c226fdc4fa523bf477e1228fe05505/lib/active_merchant/net_http_ssl_connection.rb#L6).

Previously, they actually had to work around the fact that WebMock was not assigning `@socket` (and therefore not behaving like Net::HTTP) in the past: https://github.com/activemerchant/active_merchant/commit/79c46f7ae1e33c24998f2d021888b64c22ef4e4e.

I do think WebMock should act like Net::HTTP as much as possible. Therefore, I think the solution here is to add more methods to the StubIO, which as you said @bblimke, has been around forever.
* For HTTP connections, `@socket.io` should act like a [TCPSocket](https://ruby-doc.org/stdlib-3.1.0/libdoc/socket/rdoc/TCPSocket.html).
* For HTTPS connections, `@socket.io` should act like a [OpenSSL::SSL::SSLSocket](https://ruby-doc.org/stdlib-3.1.0/libdoc/openssl/rdoc/OpenSSL/SSL/SSLSocket.html).

jdufresne:
Here is a minified test case that can be used to dig deeper. The following script works with 3.15 but not 3.16.

```rb
require 'activemerchant'
require 'webmock'
include WebMock::API

WebMock.enable!

stub_request(:post, "https://api2.authorize.net/xml/v1/request.api")

ActiveMerchant::Billing::Base.mode = 'test'

credit_card = ActiveMerchant::Billing::CreditCard.new({
  number: '4111111111111111',
  month: '11',
  year: '25',
  first_name: 'John',
  last_name: 'Doe',
  verification_value: '111',
})

gateway = ActiveMerchant::Billing::AuthorizeNetGateway.new(
  login: 'test',
  password: 'test',
)

gateway.authorize(1000, credit_card)
```

bblimke:
@rzane thank you for the investigation and explanation.

@jdufresne thank you for the code sample.

Extending StubIO to the following form works:

````
class StubIO
    def setsockopt(*args); end
    def peer_cert; end
    def ssl_version; end
    def cipher; []; end
  end
````

I'm not sure this is the right approach though. According to OpenSSL docs,
cipher should either return nil or an array of size 4.

I think the right approach should be returning nil, since there is no cipher,
yet activemerchant compatibility  patch https://github.com/activemerchant/active_merchant/commit/79c46f7ae1e33c24998f2d021888b64c22ef4e4e expects cipher to be an Array in case the socket it present.
Perhaps StubIO#cipher should return nil and activemerchant should make the change to not 
expect cipher to be an array, but allow it to be nil.

**Comment by pboling** [comment]
Seems like this might be an opportunity to also fix https://github.com/bblimke/webmock/issues/50

```
    def peer_cert; end
```

It would just need to return a stubbed cert instead of `nil`.

**Comment by bblimke** [comment]
@pboling thank you for raising that. I'm keen to apply a fix. Any thoughts on how to create a stubbed peer_cert that is going to work for most cases?

**Comment by pboling** [comment]
I was wrong about needing a stubbed cert (unless supporting old versions of `net-http` is a priority.

https://github.com/ruby/net-http/blob/e3285bd57a53b19be432c80790c63c8cde0ad6c1/lib/net/http.rb#L1596C1-L1605C1

```
    # Returns the X509 certificate chain (an array of strings)
    # for the session's socket peer,
    # or +nil+ if none.
    def peer_cert
      if not use_ssl? or not @socket
        return nil
      end
      @socket.io.peer_cert
    end
```

Modern `net-http` just returns an array of strings, or nil.

PR #978, opened 2022-06-01
Allow matching urls w/ default port and path

### LINKED ISSUE #977: `allow` doesn't match when url is https and includes path, because Addressable::URI#to_s includes port in the string
When `WebMock.disallow_net_connect!` is called with:
```rb
WebMock.disallow_net_connect!(allow: ['https://github.com/mozilla/geckodriver/releases/latest'])
```

it fails to allow requests to that url because the uri that comes into `net_connect_explicit_allowed?` is an `Addressable::URI`, who's `#to_s` method returns a string with the port in.

For example:
```rb
require 'webmock'
allowed = "https://github.com/mozilla/geckodriver/releases/latest"
uri = WebMock::Util::URI.normalize_uri(allowed)
puts uri.class
# Addressable::URI
puts uri.to_s
# https://github.com:443/mozilla/geckodriver/releases/latest

puts allowed == uri.to_s ||
      allowed == uri.host ||
      allowed == "#{uri.host}:#{uri.port}" ||
      allowed == "#{uri.scheme}://#{uri.host}:#{uri.port}" ||
      allowed == "#{uri.scheme}://#{uri.host}" && uri.port == uri.default_port
# false
```

FYI such requests happen automatically in capybara/selenium-webdriver/webdrivers when using firefox driver.

henrahmagix:
Fix #977

### DIFF
--- a/lib/webmock/webmock.rb
+++ b/lib/webmock/webmock.rb
@@ -82,7 +82,7 @@ module WebMock
       allowed == uri.host ||
       allowed == "#{uri.host}:#{uri.port}" ||
       allowed == "#{uri.scheme}://#{uri.host}:#{uri.port}" ||
-      allowed == "#{uri.scheme}://#{uri.host}" && uri.port == uri.default_port
+      (allowed == uri.omit(:port).to_s && uri.port == uri.default_port)
     else
       if allowed.respond_to?(:call)
         allowed.call(uri)


bblimke:
hi @henrahmagix 

Thank you for the PR.

It would be good to add a spec to handle that scenario.

While there is no spec for that, I believe this change can break some integrations.

E.g. someone sets allowed "http://example.com" and the request URI is "http://example.com:80/foobar"

The existing code will match and allow network connection but with the introduced change,
it now expects allowed pattern to include the path as well, therefore if this solution was to handle paths,
it would require an additional condition, rather than replacing the current one.

btw. the allowed pattern was created for just scheme host and port, since only these
are used for the actual connection. The path is only used for the HTTP requests
after the connection has already been made.
Though in most cases, we know the URL at the time of making the request
and creating the connection, therefore I think this is acceptable to allow including path
in allowed patterns.

Issue #983, opened 2022-07-19
Top-Level array in JSON body


cedrics:
Hey, we are calling an external API that expects the JSON body to be an array of hashes. It was a bit confusing to us that this does not work. 

Would you be interessted in a PR that extends the `BodyPattern` to support top level arrays? So make this test pass:

```ruby
it "should match if the request body has a top level array" do
  expect(WebMock::RequestPattern.new(:post, 'www.example.com', body: [{a: 1}])).
    to match(WebMock::RequestSignature.new(:post, "www.example.com",
    headers: {content_type: content_type}, body: "[{\"a\":1}]"))
end
```

Thanks for the great work on `webmock`. Love the library ❤️

bblimke:
Hi @cedrics .

Yes, I'd definitely be keen to make it work. I'm surprised as well that it's not supported, that nobody raised that over the last 11-12 years ;)

PR #986, opened 2022-08-03
add support for top level array in request pattern

cedrics:
Fixes https://github.com/bblimke/webmock/issues/983


bblimke:
@cedrics thank you! I appreciate the refactoring. much cleaner :)

aimee-ault:
Hi! This has a syntax error in it causing the latest release to break:
https://github.com/bblimke/webmock/blob/e060101698cba54069af480a77702290cec70618/lib/webmock/request_pattern.rb#L362

Note the typo on `reutrn` vs `return`

cedrics:
Ups, sorry 🙈 Actually that case is not covered in the tests I wrote and it is valid ruby. My bad 😭

PR #988, opened 2022-08-05
Make the StubSocket slightly more believable

rzane:
Fixes https://github.com/bblimke/webmock/pull/976#issuecomment-1203449899

* ActiveMerchant expects `#ssl_version` and `#cipher` to be defined. Though these methods are private, it's best if webmock can *try* to act like the Net::HTTP to avoid gems having to detect if WebMock is enabled.
* I also took the opportunity to make `StubSocket#close` and `StubSocket#closed?` a bit more realistic as well.
* I also noticed that `Net::HTTP#ipaddr` was broken because `StubSocket#peeraddr` wasn't defined.

### DIFF
--- a/lib/webmock/http_lib_adapters/net_http.rb
+++ b/lib/webmock/http_lib_adapters/net_http.rb
@@ -128,7 +128,11 @@ module WebMock
 
 
         def ensure_actual_connection
-          do_start if @socket.is_a?(StubSocket)
+          if @socket.is_a?(StubSocket)
+            @socket&.close
+            @socket = nil
+            do_start
+          end
         end
 
         alias_method :start_with_connect, :start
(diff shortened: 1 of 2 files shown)


bblimke:
Thank you @rzane . Awesome!

lib/webmock/http_lib_adapters/net_http.rb
```diff
@@ -128,7 +128,11 @@ def start_without_connect
 
 
         def ensure_actual_connection
-          do_start if @socket.is_a?(StubSocket)
+          if @socket.is_a?(StubSocket)
+            @socket&.close
+            @socket = nil
```
ElvinEfendi, line 133: Why are these necessary? How is it better simulating net/http?

It causes issues when running tests with tools like https://github.com/shopify/toxiproxy where the test involves starting a successful HTTP connection to a local server, and then mimics that server failing using toxiproxy, and attempting another request. The exception raised is "undefined method `closed?' for nil", which is coming from https://github.com/ruby/net-http/blob/d39f1e35da0094759356a49a3c96fea84325c388/lib/net/http.rb#L2431.

PR #989, opened 2022-08-05
Requires Ruby 2.3+

sue445:
https://github.com/bblimke/webmock/pull/969 uses `&.`, but `&.` has been available since Ruby 2.3
c.f. https://www.ruby-lang.org/en/news/2015/12/25/ruby-2-3-0-released/

webmock v3.15.0+ already doesn't work with Ruby < 2.3, so I fixed `required_ruby_version`


bblimke:
Thank you @sue445

PR #990, opened 2022-08-07
Remove dead code for supporting Ruby < 2.3

rzane:
As of #989, this library requires Ruby >= 2.3, so the following changes were in order:

* The `validate_headers` function is only executed for Ruby versions prior to 2.3.
* The patron adapter had tests that were only run for Ruby 1.9. These tests should actually have been written for all Ruby versions _greater than_ 1.9.


bblimke:
Great cleanup. Thank you @rzane !

PR #991, opened 2022-08-07
Avoid replacing Net::BufferedIO

rzane:
Ruby's Net::BufferedIO doesn't hit the network; therefore, this library should not be patching it.

I believe that `WebMockNetBufferedIO` and `PatchedIO` were introduced so that a String could be used as a raw response. However; it's actually much simpler just to convert the string to a StringIO before attempting to parse the raw HTTP response.


bblimke:
Thank you @rzane . That is a solid simplification of Net::HTTP adapter. Interesting that introduction of custome BufferedIO led to so much added complexity before.

PR #992, opened 2022-08-08
Simplify Net::HTTP connection handing

rzane:
This PR aims to simplify the way that the stub socket is assigned and swapped in Net::HTTP.

### DIFF
--- a/lib/webmock/http_lib_adapters/net_http.rb
+++ b/lib/webmock/http_lib_adapters/net_http.rb
@@ -77,7 +77,6 @@ module WebMock
           WebMock::RequestRegistry.instance.requested_signatures.put(request_signature)
 
           if webmock_response = WebMock::StubRegistry.instance.response_for_request(request_signature)
-            @socket = Net::HTTP.socket_type.new
             WebMock::CallbackRegistry.invoke_callbacks(
               {lib: :net_http}, request_signature, webmock_response)
             build_net_http_response(webmock_response, &block)
--- a/lib/webmock/http_lib_adapters/net_http.rb
+++ b/lib/webmock/http_lib_adapters/net_http.rb
@@ -72,6 +72,8 @@ module WebMock
         end
 
         def request(request, body = nil, &block)
+          return super unless started?
+
           request_signature = WebMock::NetHTTPUtility.request_signature_from_request(self, request, body)
 
           WebMock::RequestRegistry.instance.requested_signatures.put(request_signature)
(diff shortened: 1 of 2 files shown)


lib/webmock/http_lib_adapters/net_http.rb
```diff
@@ -72,78 +72,57 @@ def constants(inherit=true)
         end
 
         def request(request, body = nil, &block)
+          return super unless started?
```
rzane, line 75: This lets us avoid having to deal with the case where the connection is not started yet.

see https://github.com/ruby/net-http/blob/196f3d70f9ce154dc63a86a2cdbf1aec4df4982e/lib/net/http.rb#L1568-L1574

lib/webmock/http_lib_adapters/net_http.rb
```diff
@@ -72,78 +72,57 @@ def constants(inherit=true)
...
+
           request_signature = WebMock::NetHTTPUtility.request_signature_from_request(self, request, body)
 
           WebMock::RequestRegistry.instance.requested_signatures.put(request_signature)
 
           if webmock_response = WebMock::StubRegistry.instance.response_for_request(request_signature)
-            @socket = Net::HTTP.socket_type.new
```
rzane, line 80: There's no reason to replace the socket here. At the very least, we'd want to close the old socket before initializing a new one.

lib/webmock/http_lib_adapters/net_http.rb
```diff
@@ -72,78 +72,57 @@ def constants(inherit=true)
...
-            @socket&.close
-            @socket = nil
-            do_start
-          end
-        end
-
-        alias_method :start_with_connect, :start
-
-        def start(&block)
+        def connect
```
rzane, line 111: `Net::HTTP#start` calls `#connect`. This is a much simpler place for us to hook in, because unlike `#start`, `#connect` doesn't have a block variant.

spec/acceptance/net_http/net_http_spec.rb
```diff
@@ -215,14 +215,14 @@ class TestMarshalingInWebMockNetHTTP
 
   if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.7.0')
     it "uses the StubSocket to provide IP address" do
-      Net::HTTP.start("http://example.com") do |http|
+      Net::HTTP.start("example.com") do |http|
```
rzane, line 218: This is me fixing a mistake I made in #988 😄

lib/webmock/http_lib_adapters/net_http.rb
bblimke: Clever! 👍

lib/webmock/http_lib_adapters/net_http.rb
bblimke: 👍

bblimke:
Thank you @rzane . Much simpler.

PR #994, opened 2022-08-08
Fix Syntax Error: `reutrn`

mspangler:
We have a syntax error, `reutrn`, when matching on an array body.


aimee-ault (approved):

bblimke:
@mspangler thank you!

Linuus:
Thanks for the fix! I was just going to submit my own when I noticed it was fixed already :)

Is it possible to make a patch release?

PR #1005, opened 2022-10-28
em-http-request: do not alter non stubbed requests

ylecuyer:
Following this change: https://github.com/bblimke/webmock/pull/936

We are now altering the header that are sent when using webmock and em-http-request in `WebMock.allow_net_connect!` mode

If you check here https://github.com/igrigorik/em-http-request/blob/34919d760b940dfd27159bbebce5c4dac9845eb6/lib/em-http/client.rb#L180 when forwarding the request to em-http-request, the `send_request` method expects some headers to be downcased (otherwise they are doubled)

This is an issue for example for `Content-Length` which would be interpreted as an array in the receiving webserver and not as a number.

The idea of the fix is to keep the raw_headers (not normalized by webmock) and use those raw_headers when forwarding the request

### DIFF
--- a/lib/webmock/http_lib_adapters/em_http_request_adapter.rb
+++ b/lib/webmock/http_lib_adapters/em_http_request_adapter.rb
@@ -116,7 +116,7 @@ if defined?(EventMachine::HttpClient)
           }
           self
         elsif WebMock.net_connect_allowed?(request_signature.uri)
-          super
+          super(request_signature.raw_headers, request_signature.body)
         else
           raise WebMock::NetConnectNotAllowedError.new(request_signature)
         end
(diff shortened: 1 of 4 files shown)


bblimke:
Thank you for raising the issue and offering a solution @ylecuyer . While your solution works, the RequestSignature object should only be responsible for handing required information for matching the request and should not be used to carry additional data required for adapter implementation. Therefore I have introduced an alternative solution that still makes your spec pass: https://github.com/bblimke/webmock/commit/ef566e82a2ebf378a4033e401838625c1adebc8c and all the changes stay inside the em_http_request adapter.

ylecuyer:
Thank you !

PR #1006, opened 2022-11-08
Version checker fails under some conditions

marcrohloff:
For ex we are using Curb version 0.9.7 and it prints a warning even though it shouldn't


tubaxenor:
Any chance this or https://github.com/bblimke/webmock/pull/1024 can be merged? It's been up for a while 😢

bblimke:
Thank you @marcrohloff for the complete solution with many more cases testes 👍

PR #1012, opened 2022-12-12
Json-ify any non-string object in #to_return_json

inkstak:
When using `to_return_json`, it allows to pass an Array body.
By extension, it'll convert any non-string object passed to `:body` parameter to json. 

```ruby
stub_request(...).to_return_json(body: [{abc: "def" }])
stub_request(...).to_return_json(body: Record.find(1))
```


jasonkarns:
Love this!

Any chance it can be extended to accept procs also? It's a jarring inconsistency that `to_return` can accept a `body: -> {}` lambda but `to_return_json` doesn't.

Rather disappointing that in order for the stub to trigger any side effects, we have to drop out of `to_return_json` and lose the header/body niceties.

inkstak:
Added to the PR: 

```ruby
stub_request(...).to_return_json(body: ->{ Record.find(1) })
```

inkstak:
@bblimke Thank you for maintaining this awesome gem !

PR #1013, opened 2022-12-14
Fix matching stubs with HashExcludingMatcher

lucasarnaud:
Stubs using HashExcludingMatcher were not working because body is not being converted to a hash in order to do the matching


bblimke (approved):

bblimke:
Thank you @lucasarnaud !

PR #1024, opened 2023-04-24
Fix version check when major version is less than the max version

romanstrazanec:
This should fix the error below

> You are using Curb 0.9.11. WebMock is known to work with Curb >= 0.7.16, < 1.1, except versions 0.8.7. It may not work with this version.


bblimke:
Thank you @romanstrazanec

Issue #1027, opened 2023-05-18
Performance of `WebMock::Util::Headers.normalize_headers`


baweaver:
While trying to sort out some Capybara speed issues we were running memory profiling and noticed that there was an exceptionally hot path in the `normalize_headers` method. We took a more detailed look at it and came up with this experiment to see what interesting information we could find on it:

```ruby
require 'benchmark/ips'
require 'memory_profiler'

# Recreating the Header utils
module WebMock
  module Util
    class Headers
      # Hoist these so we don't end up with recreated objects
      STANDARD_HEADER_DELIMITER = '-'.freeze
      NONSTANDARD_HEADER_DELIMITER = '_'.freeze
      JOIN = ', '.freeze

      # What exists today, unchanged
      def self.normalize_headers(headers)
        return nil unless headers
        array = headers.map { |name, value|
          [name.to_s.split(/_|-/).map { |segment| segment.capitalize }.join("-"),
           case value
            when Regexp then value
            when Array then (value.size == 1) ? value.first.to_s : value.map {|v| v.to_s}.sort
            else value.to_s
           end
          ]
        }
        Hash[*array.inject([]) {|r,x| r + x}]
      end

      def self.just_return_headers(headers)
        headers
      end

      # Experiment to try and speed that up
      def self.enhanced_normalize_headers(headers)
        return nil unless headers

        # Using `each_with_object` to reduce object creations and directly build
        # the new headers hash.
        headers.each_with_object({}) do |(name, value), new_headers|
          # Moved the name normalization into a new method mostly to give a name to what it's
          # doing here.
          new_headers[normalize_name(name)] =
            case value
            when Regexp then value
            when Array then (value.size == 1) ? value.first.to_s : value.map(&:to_s).sort
            else value.to_s
            end
        end
      end

      # Good deal of time savings on that process though
      def self.normalize_name(name)
        name
          .to_s
          # By using `tr` here we can normalize the headers ahead of time
          .tr(NONSTANDARD_HEADER_DELIMITER, STANDARD_HEADER_DELIMITER)
          # ...which will be faster because `split` on a `String` is a good deal faster
          # than on a `Regexp`.
          .split(STANDARD_HEADER_DELIMITER)
          # Now since this is state we control here we can mutate it with `map!` to again
          # avoid extra object creations.
          .map!(&:capitalize)
          # Then join it back together. Most of these constants are to avoid recreating
          # these strings hundreds of thousands of times in larger codebases. Ours I
[...]

PR #1029, opened 2023-06-06
Optimizes `normalize_headers` for performance

baweaver:
Optimizes the `WebMock::Util::Headers.normalize_headers` method to:

* Create less objects
* Avoid using Regexp where possible
* Hoists constants

More details can be found in the related issue, including performance benchmarks:

https://github.com/bblimke/webmock/issues/1027

### DIFF
--- a/lib/webmock/util/headers.rb
+++ b/lib/webmock/util/headers.rb
@@ -6,18 +6,21 @@ module WebMock
 
     class Headers
 
+      STANDARD_HEADER_DELIMITER = '-'.freeze
+      NONSTANDARD_HEADER_DELIMITER = '_'.freeze
+      JOIN = ', '.freeze
+
       def self.normalize_headers(headers)
         return nil unless headers
-        array = headers.map { |name, value|
-          [name.to_s.split(/_|-/).map { |segment| segment.capitalize }.join("-"),
-           case value
+
+        headers.each_with_object({}) do |(name, value), new_headers|
+          new_headers[normalize_name(name)] =
+            case value
             when Regexp then value
-            when Array then (value.size == 1) ? value.first.to_s : value.map {|v| v.to_s}.sort
+            when Array then (value.size == 1) ? value.first.to_s : value.map(&:to_s).sort
             else value.to_s
-           end
-          ]
-        }
-        Hash[*array.inject([]) {|r,x| r + x}]
+            end
+        end
       end
 
       def self.sorted_headers_string(headers)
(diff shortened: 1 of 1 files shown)


lib/webmock/util/headers.rb
```diff
@@ -57,6 +60,23 @@ def self.basic_auth_header(*credentials)
...
+          # ...which will be faster because `split` on a `String` is a good deal faster
+          # than on a `Regexp`.
+          .split(STANDARD_HEADER_DELIMITER)
+          # Now since this is state we control here we can mutate it with `map!` to again
+          # avoid extra object creations.
+          .map!(&:capitalize)
+          # Then join it back together. Most of these constants are to avoid recreating
+          # these strings hundreds of thousands of times in larger codebases. Ours I
+          # believe was in the millions.
+          .join(STANDARD_HEADER_DELIMITER)
```
baweaver, line 77: I may remove these comments real quick, as they don't add a lot of value to the library itself.

lib/webmock/util/headers.rb
```diff
@@ -6,18 +6,21 @@ module Util
...
+      NONSTANDARD_HEADER_DELIMITER = '_'.freeze
+      JOIN = ', '.freeze
+
       def self.normalize_headers(headers)
         return nil unless headers
-        array = headers.map { |name, value|
-          [name.to_s.split(/_|-/).map { |segment| segment.capitalize }.join("-"),
-           case value
+
+        headers.each_with_object({}) do |(name, value), new_headers|
```
natematykiewicz, line 16: I'm not a maintainer, but `each_with_object` is an ActiveSupport method, and webmock doesn't depend on ActiveSupport. You'll probably need to change this for the test suite to pass.

natematykiewicz: Wow. TIL. Thanks for the clarification!

technicalpickles:
Is there anything that can be done to help this along? I've been profiling our app's system specs (same one as @baweaver), and this is one I keep coming back to.

lib/webmock/util/headers.rb
```diff
@@ -57,6 +60,15 @@ def self.basic_auth_header(*credentials)
         "Basic #{Base64.strict_encode64(credentials.join(':')).chomp}"
       end
 
+      def self.normalize_name(name)
```
bblimke, line 63: This code is very easy to read now. Thank you @baweaver !

technicalpickles:
Thank you for merging 🙇🏻  Any chance we can get a release with it and https://github.com/bblimke/webmock/pull/1033 🙏🏻

bblimke:
@technicalpickles 3.19.0 is now released. thank you!

PR #1030, opened 2023-06-12
add "[::1]" for localhost determination

Yuki-Inoue:
Currently, `allow_localhost` option of `disable_net_connect!` method does not allow `[::1]` to be communicated, while this is actually an localhost address (of IPv6).

This PR should allow communicating with `http://[::1]/some/path` address.


zenspider:
I just added exactly the same patch locally.

PR #1038, opened 2023-10-14
Set Typhoeus timing fields when constructing Response objects

Roguelazer:
This sets all of the `*_time` fields from [Typhoeus::Response::Informations](https://www.rubydoc.info/github/typhoeus/typhoeus/Typhoeus/Response/Informations) to 0.0 when constructing the `Typhoeus::Response` objects.

It would be cool to have an API for mocking these; perhaps something like
```ruby
stub_request(:get, "http://www.example.com")
  .to_return(status: 200)
  .to_take(3.seconds)
```

but for now I just wanted to set them to a non-nil value


bblimke:
@Roguelazer thank you for implementing this change. Can you confirm that all these attributes are always not nil for real requests? If yes, then your change looks perfect.

PR #1040, opened 2023-10-24
Fix failing Async::HTTP specs due to Async API change

### LINKED ISSUE #1036: CI consistently failing on Ruby >= 3.1, is it ok to update?
Hi,

given the broken status badge and the results at 

https://github.com/bblimke/webmock/actions/runs/6021150216/

It is not clear if the current version is supposed to work with Ruby >= 3.1

I did upgrade to 3.19 and webmock is still working, but I would ask for a confirmation

mattbrictson:
The `async` gem recently made a breaking API change that was released in patch version 2.6.4.[^1] Following the change, async tasks that return an  exception object will no longer raise that exception when `wait` is called.

The `Async::HTTP` specs in webmock were written to leverage the old behavior of `wait`. Since async 2.6.4 was released, these specs have been failing in CI.

This commit fixes the failing specs by updating how `wait` is used, such that exceptions are still raised as expected in async 2.6.4. This should restore CI to a working state.

Fixes #1036 

[^1]: https://github.com/socketry/async/pull/270

### DIFF
--- a/spec/acceptance/async_http_client/async_http_client_spec.rb
+++ b/spec/acceptance/async_http_client/async_http_client_spec.rb
@@ -345,7 +345,7 @@ unless RUBY_PLATFORM =~ /java/
     end
 
     def make_request(method, url, protocol: nil, headers: {}, body: nil)
-      Async do
+      result  = Async do
         endpoint = Async::HTTP::Endpoint.parse(url)
 
         begin
@@ -362,6 +362,8 @@ unless RUBY_PLATFORM =~ /java/
           e
         end
       end.wait
+
+      result.is_a?(Exception) ? raise(result) : result
     end
 
     def response_to_hash(response)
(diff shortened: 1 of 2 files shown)


spec/acceptance/async_http_client/async_http_client_spec.rb
```diff
@@ -362,6 +362,8 @@ def make_request(method, url, protocol: nil, headers: {}, body: nil)
           e
         end
       end.wait
+
+      result.is_a?(Exception) ? raise(result) : result
```
mattbrictson, line 366: 🗒️ Before async 2.6.4, `wait` would raise an exception if the task returned an exception object (in this case, `Async::TimeoutError`). Since 2.6.4, `wait` no longer raises the exception, it just returns it as an exception object.

The webmock specs rely on the exception being raised, so I added this ternary line to restore the old `wait` behavior.

mattbrictson:
Looks like this PR is restoring Ruby 3.1 and 3.2 to green status in CI. 🎉 

As a separate issue, Ruby head is failing due to the `cookiejar` gem (which is no longer being maintained) not supporting Ruby 3.3. It is a dependency of `em-http-request` and there is already an open issue for it: https://github.com/igrigorik/em-http-request/issues/354.

bblimke:
@mattbrictson big thanks for the thorough investigation and the fix! 🙏

PR #1042, opened 2023-10-27
EmHttpRequestAdapter - Call detect instead of select..first

RicardoTrindade:
Tiny tiny optimization PR

As per https://github.com/fastruby/fast-ruby/blob/main/code/enumerable/select-first-vs-detect.rb
Detect is slightly faster than select{}.first


bblimke:
Thank you for this optimisation @RicardoTrindade 👍

PR #1041, opened 2023-10-26
Add `base64` to runtime dependency

koic:
This PR adds `base64` to runtime dependency to suppress the following Ruby 3.3's warning:

```console
$ ruby -v
ruby 3.3.0dev (2023-10-23T08:04:27Z master e6fcf07a6f) [x86_64-darwin22]

$ path/to/rubocop
$ bundle exec rspec
/Users/koic/.rbenv/versions/3.3.0-dev/lib/ruby/gems/3.3.0+0/gems/webmock-3.19.1/lib/webmock/util/headers.rb:3:
warning: base64 which will no longer be part of the default gems since Ruby 3.4.0. Add base64 to your Gemfile or gemspec.
```

I found this warning in the RuboCop repository:
https://app.circleci.com/pipelines/github/rubocop/rubocop/10108/workflows/20097132-bbe3-4fa4-ad7e-de7c30c86e69/jobs/291956?invite=true#step-104-0_163

cf: https://github.com/rails/rails/pull/48907

PR #1046, opened 2023-12-23
Remove runtime dependency on `base64`

Earlopain:
Ruby 3.3 shows a warning when using base64 without declaring it in the gemspec. Ruby 3.4 will remove base64 as a default gem.

Since this gem is only using a single method in one place there is no need to pull in the full dependency. It is only a wrapper over `Array#pack`.

Alternative to #1041

Other gems have taken the same/similar approach:
* https://github.com/rubocop/rubocop/pull/12313
* https://github.com/rack/rack/pull/2110
* https://github.com/newrelic/newrelic-ruby-agent/pull/2378
* https://github.com/lostisland/faraday/pull/1541 (pending)

### DIFF
--- a/lib/webmock/util/headers.rb
+++ b/lib/webmock/util/headers.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'base64'
-
 module WebMock
 
   module Util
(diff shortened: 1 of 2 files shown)


mattbrictson:
As of [2 weeks ago](https://github.com/ruby/ruby/pull/9550), it looks like Ruby head has now removed the default base64 gem entirely. So projects that use webmock fail with this exception:

> /home/runner/.rubies/ruby-head/lib/ruby/3.4.0+0/bundled_gems.rb:74:in `require': cannot load such file -- base64 (LoadError)

A workaround exists: any project using webmock also has to add base64 to their Gemfile. Still, it would be nice to see this PR merged. Thanks! 🙏

bblimke:
@Earlopain thank you for providing this solution. I appreciate the simplicity! 👍 It's great to have a solution that doesn't require adding another dependency to the gemspec.

PR #1047, opened 2024-01-24
Remove redundant freeze

znz:
since frozen_string_literal magic comment added.


bblimke:
Thank you, @znz! 👍 Great catch on those redundant string freeze statements.

PR #1049, opened 2024-02-13
Don't use deprecated Rack::VERSION in Rack 3

etiennebarrie:
When I upgraded Rack, I got a few deprecation warnings regarding the usage of `Rack::VERSION` in `lib/webmock/rack_response.rb`.

This conditionally adds the key to the Rack environment, since not only it's [no longer required](https://github.com/rack/rack/blob/v3.0.1/UPGRADE-GUIDE.md#rackversion-is-no-longer-required) but it's also [deprecated](https://github.com/rack/rack/blob/v3.0.1/lib/rack/version.rb#L17).

I also added a commit to enable deprecation warnings by default, it's [not something that RSpec does](https://github.com/rspec/rspec-core/issues/2867) but it can be useful like in this case. Let me know (or just remove that commit) if you'd rather not have that.

### DIFF
--- a/lib/webmock/rack_response.rb
+++ b/lib/webmock/rack_response.rb
@@ -47,7 +47,9 @@ module WebMock
       # Rack-specific variables
       env['rack.input']      = StringIO.new(body)
       env['rack.errors']     = $stderr
-      env['rack.version']    = Rack::VERSION
+      if !Rack.const_defined?(:RELEASE) || Rack::RELEASE < "3"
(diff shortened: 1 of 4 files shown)


bblimke:
Thank you @etiennebarrie . I appreciate the fix and enabling deprecation warning by default in webmock specs.

etiennebarrie:
Thanks for releasing a new version this quickly!

PR #1050, opened 2024-02-13
httpclient_adapter: use uri after filters when building request signature

mattbnz:
If a filter changes the URI (e.g. to enforce a protocol or change the hostname of the request) this is currently unable to be mocked as webmock fetches the URI for the signature of the request before calling the filters!

### DIFF
--- a/lib/webmock/http_lib_adapters/httpclient_adapter.rb
+++ b/lib/webmock/http_lib_adapters/httpclient_adapter.rb
@@ -157,13 +157,13 @@ if defined?(::HTTPClient)
     end
 
     def build_request_signature(req, reuse_existing = false)
-      uri = WebMock::Util::URI.heuristic_parse(req.header.request_uri.to_s)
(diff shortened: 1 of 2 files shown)


bblimke:
@mattbnz thank you for submitting the PR. you are correct that one expects the signature to be based on the final url, after all filters applied. Would you care to add at least one test to https://github.com/bblimke/webmock/blob/master/spec/acceptance/httpclient/httpclient_spec.rb to verify that the behaviour works as expected?
Really appreciate your help on this!

mattbnz:
@bblimke thank you for the great library and quick response! 

I'm pretty new to ruby and rspec so any guidance on the desired test style is welcome - for now I just modified the expectations of the existing tests to include URI modification in the filter - seemed to me straightforward enough and simple enough in concept to the existing tests that it made sense vs creating a new spec/test overall... but if you prefer a completely separate test let me know.

Verified that the changed spec fails on master and passes in this branch.

Cheers

bblimke:
@mattbnz thank you. I believe the provided extension of existing spec is sufficient 👍

Issue #1055, opened 2024-04-26
async-http adapter has an implicit dependency on async-io which was removed from async-http at version 0.65.0


jakeonfire:
as of https://github.com/socketry/async-http/releases/tag/v0.65.0 `async-http` does not use `async-io` anymore. but the `async-http` webmock adapter has an implicit dependency on `async-io` here: https://github.com/bblimke/webmock/blob/fc6a2ab897a069d861adbc1c88e51b2cf8aa88ac/lib/webmock/http_lib_adapters/async_http_client_adapter.rb#L156-L160

jakeonfire:
cc @ayanko

jakeonfire:
https://github.com/bblimke/webmock/pull/1056

PR #1056, opened 2024-04-26
Replace Async::IO::Socket usage with stdlib Socket…

jakeonfire:
…for `async-http` adapter to remove implicit dependency on `async-io`

`async-io` was removed as a dependency from `async-http` in version 0.65.0: https://github.com/socketry/async-http/releases/tag/v0.65.0

fixes: https://github.com/bblimke/webmock/issues/1055

### DIFF
--- a/lib/webmock/http_lib_adapters/async_http_client_adapter.rb
+++ b/lib/webmock/http_lib_adapters/async_http_client_adapter.rb
@@ -152,11 +152,15 @@ if defined?(Async::HTTP)
 
         private
 
+        def socket_class
+          @_socket_class ||= Gem::Dependency.new("async-io").matching_specs.any? ? Async::IO::Socket : Socket
+        end
+
         def create_connected_sockets
           pair = begin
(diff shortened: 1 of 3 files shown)


jpalermo:
@bblimke, could you take a look at this when you get a chance?

bblimke:
@jakeonfire Thank you for the PR and the fix. Is it worth having conditional logic to keep compatibility with version < 0.65?

jakeonfire:
these changes are backwards-compatible

jakeonfire:
should i revert the version number change? otherwise you can overwrite it after merging.

bblimke:
> these changes are backwards-compatible

The Ruby 3.0 build, which uses async-http 0.64.2 is failing.

jakeonfire:
@bblimke looks good now 👍

bblimke:
Much simpler indeed. Thank you @jakeonfire 👍

bblimke:
It's now released as version 3.23.1

PR #1065, opened 2024-08-01
Mock HTTP::Connection#finished_request?

c960657:
Mock [`HTTP::Connection#finished_request?`](https://github.com/httprb/http/blob/b74b16cd55381433627205c90bc126f9b81600ff/lib/http/connection.rb#L139C9-L141) in http.rb.


c960657:
The tests currently fail due to https://github.com/bblimke/webmock/pull/1066 and https://github.com/igrigorik/em-http-request/pull/365.

bblimke:
The solution looks great. Thank you @c960657 !

PR #1066, opened 2024-08-02
Ignore parse exceptions

c960657:
Tests are currently failing in `master` due to a change in the latest version of REXML.

REXML previously accepted some invalid XML (e.g. `foo bar`) but raised on other (`<foo bar`). Since [v3.3.3](https://github.com/ruby/rexml/releases/tag/v3.3.3) it also raises on `foo bar` (see https://github.com/ruby/rexml/pull/184).

Likewise, `WebMock::Util::JSON.parse` accepts some invalid JSON (e.g. `foo bar`) but raises on other (e.g. `[foo bar`).

It seems both matchers should just rescue and ignore parse exceptions.

### DIFF
--- a/lib/webmock/request_pattern.rb
+++ b/lib/webmock/request_pattern.rb
@@ -309,6 +309,7 @@ module WebMock
       else
         WebMock::Util::QueryMapper.query_to_values(body, notation: Config.instance.query_values_notation)
       end
+    rescue Psych::SyntaxError, REXML::ParseException
     end
 
     def body_format(content_type)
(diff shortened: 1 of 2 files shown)


c960657:
The Ruby head test fails due to https://github.com/igrigorik/em-http-request/pull/365.

bquorning (approved):

bblimke:
Thank you @c960657 ! I appreciate the explanation and the PR. I agree it makes sense to ignore the parsing exceptions and simply not match the request body. 

Any suggestions on how to fix the em-http-client issue in Ruby head, or just wait for em-http-client to be fixed?

c960657:
Hmm, no, I don't think there is anything WebMock can do about the em-http-client issue.

bblimke:
It is annoying that webmock build fails due to a problem with just one of http clients.

c960657:
We can use `skip` to bypass the tests for em-http-client on Ruby HEAD (see #1070).

PR #1068, opened 2024-09-10
Use Patron::Session instead of Patron namespace

### LINKED ISSUE #1059: Patron name clash with business entity
If an application has a table `patrons` and a model Patron, then it causes webmock not to work, because it checks if Patron is defined. Because Patron is defined, it fails with `uninitialized constant Patron::Session`

```
# lib/webmock/http_lib_adapters/patron_adapter.rb
begin
  require 'patron'
rescue LoadError
  # patron not found
end

if defined?(::Patron) # <<<--- replace this?
```

rodrigoargumedo:
Closes #1059


bblimke:
Thank you @rodrigoargumedo

PR #1069, opened 2024-09-14
FEATURE: Allow asserting on requests in tests

johngallagher:
## Context

@jamesshore has created the concept of ["Testing With Nullables"](https://www.jamesshore.com/v2/projects/nullables/a-light-introduction-to-nullables).

>  I’ve figured out another way. A way that doesn’t use end-to-end tests, doesn’t use mocks, doesn’t ignore infrastructure, doesn’t require a rewrite. It’s something you can start doing today, and it gives you the speed, reliability, and maintainability of unit tests with the power of end-to-end tests.

A small snippet:

```javascript
it("reads command-line argument, transform it with ROT-13, and writes result", () => {
  const { output } = run({ args: [ "my input" ] });
  assert.deepEqual(output.data, [ "zl vachg\n" ];
});

function run({ args = [] } = {}) {
 const commandLine = CommandLine.createNull({ args });
 const output = commandLine.trackOutput();

 const app = new App(commandLine);
 app.run();

 return { output };
}
```

I've been experimenting with these techniques in various codebases and love the code that results.

## Why

This style of assertions, whilst not adhering to James' full pattern, is a thin layer on top of what Webmock is already doing - @jamesshore calls it ["Output Tracking"](https://www.jamesshore.com/v2/projects/nullables/testing-without-mocks#output-tracking)

## Example

Let's say we're doing an API request to Cloudflare to get IP blocking rules.

### Before

```
  it "makes a get request to Cloudflare to get the rules" do
    stub_request(:get, "www.cloudflare.com/api/v2/rules").to_return_json(body: [{...}])
    subject.call
    expect(WebMock).to have_requested(:get, "www.cloudflare.com/api/v2/rules").
      with(query: {"ip" => "2.5.4.3"})
  end
```

**Strengths**

* Leans into the Ruby and RSpec metaprogramming conventions
* Reads like English

**Weaknesses**

* Maintaining a `#have_requested` RSpec matcher means more code
* Custom matchers means writing readable failure messages
* RSpec magic means that if the custom matcher fails it can be difficult to debug

### After

```ruby
  it "makes a get request to Cloudflare to get the rules" do
    stub_request(:get, "www.cloudflare.com/api/v2/rules").to_return_json(body: [{...}])
    subject.call
    expect(requests_made.count).to eq(1)
    expect(requests_made.first.method).to eq(:get)
    expect(requests_made.first.uri.host).to eq("www.cloudflare.com")
    expect(requests_made.first.query).to eq(ip: "2.5.4.3")
  end
```

**Strengths**

* Plain old Ruby
* Works with Minitest with no extra code or magic needed
* Allows asserting on order of requests
* Opens up options for extra helper methods
* Reduces coupling to Webmock - we could use another mocking library, implement `#requests_made` and we'd be good

**Weaknesses**

* `#requests_made` needs a mixin to work
* Can be more verbose (this can be mitigated by writing helper methods)

### After - Cleaner

```ruby
  it "makes a get request to Cloudflare to get the rules" do
    stub_request(:get, "www.cloudflare.com/api/v2/rules").to_return_json(body: [{...}])
    subject.call
    expect(cloudflare_requests_made.count).to eq(1)
    expect(cloudflare_requests_made.first.method).to eq(:get)
    expect(cloudflare_requests_made.first.query).to eq(ip: "2.5.4.3")
  end

  # private method for more readability and resilience
  def cloudflare_requests_made
    requests_made.select { |r| r.uri.host == "www.cloudflare.com" }
  end
```

## How

* Adjust the `HashCounter` class to store requests in an array
* Add some extra methods onto the RequestSignature

### DIFF
--- a/lib/webmock/request_signature.rb
+++ b/lib/webmock/request_signature.rb
@@ -44,6 +44,10 @@ module WebMock
       !!(headers&.fetch('Content-Type', nil)&.start_with?('application/json'))
     end
 
+    def parsed_body
+      JSON.parse(body, symbolize_names: true)
+    end
+
     private
 
     def assign_options(options)
--- a/lib/webmock/util/hash_counter.rb
+++ b/lib/webmock/util/hash_counter.rb
@@ -13,19 +13,24 @@ module WebMock
         @max = 0
         @lock = ::Mutex.new
         self.array = []
+        @request_object_ids = {}
       end
 
       def put(key, num=1)
         @lock.synchronize do
-          if hash.key?(key)
-            existing_key = hash.keys.find { |k| k.hash == key.hash }
-            array << existing_key
-          else
-            array << key
-          end
+          store_to_array(key:, num:)
           hash[key] += num
           @order[key] = @max += 1
         end
+      end
+
+      def store_to_array(key:, num:)
+        request_object_id = @request_object_ids[key.hash]
+        request_object_id = key.object_id if request_object_id.nil?
+        num.times do
+          array << ObjectSpace._id2ref(request_object_id)
+        end
+        @request_object_ids[key.hash] = key.object_id
       end
 
       def get(key)
--- a/lib/webmock/util/hash_counter.rb
+++ b/lib/webmock/util/hash_counter.rb
@@ -17,9 +17,14 @@ module WebMock
 
       def put(key, num=1)
         @lock.synchronize do
+          if hash.key?(key)
+            existing_key = hash.keys.find { |k| k.hash == key.hash }
+            array << existing_key
+          else
+            array << key
+          end
           hash[key] += num
           @order[key] = @max += 1
-          array << key
         end
       end
 
--- a/lib/webmock/util/hash_counter.rb
+++ b/lib/webmock/util/hash_counter.rb
@@ -25,14 +25,14 @@ module WebMock
       end
 
       def store_to_array(key:, num:)
-        request_object_id = @request_object_ids[key.hash]
+        request_object_id = @request_object_ids[key]
         request_object_id = key.object_id if request_object_id.nil?
         num.times do
           array << ObjectSpace._id2ref(request_object_id)
         rescue RangeError
           # Points to an invalid or recycled object so ignore
         end
-        @request_object_ids[key.hash] = key.object_id
+        @request_object_ids[key] = key.object_id
       end
 
       def get(key)
--- a/lib/webmock/util/hash_counter.rb
+++ b/lib/webmock/util/hash_counter.rb
@@ -29,6 +29,8 @@ module WebMock
         request_object_id = key.object_id if request_object_id.nil?
         num.times do
           array << ObjectSpace._id2ref(request_object_id)
+        rescue RangeError
+          # Points to an invalid or recycled object so ignore
         end
         @request_object_ids[key.hash] = key.object_id
       end
--- a/lib/webmock/util/hash_counter.rb
+++ b/lib/webmock/util/hash_counter.rb
@@ -5,19 +5,21 @@ require 'thread'
 module WebMock
   module Util
     class HashCounter
-      attr_accessor :hash
+      attr_accessor :hash, :array
 
       def initialize
         self.hash = Hash.new(0)
         @order = {}
         @max = 0
         @lock = ::Mutex.new
+        self.array = []
       end
 
       def put(key, num=1)
         @lock.synchronize do
           hash[key] += num
           @order[key] = @max += 1
+          array << key
         end
       end
 
--- a/lib/webmock/request_registry.rb
+++ b/lib/webmock/request_registry.rb
@@ -7,25 +7,6 @@ module WebMock
 
     attr_accessor :requested_signatures
 
-    class Request
-      attr_accessor :method, :uri, :headers, :body
-
-      def self.from_webmock_request_signature(request_signature)
-        new(method: request_signature.method, uri: request_signature.uri, headers: request_signature.headers, body: request_signature.body)
-      end
-
-      def initialize(method:, uri:, headers:, body:)
-        @method = method
-        @uri = uri
-        @headers = headers
-        @body = body
-      end
-
-      def parsed_body
-        JSON.parse(body, symbolize_names: true)
-      end
-    end
-
     def initialize
       reset!
     end
@@ -46,8 +27,7 @@ module WebMock
 
     def to_a
       requested_signatures.
-        array.
-        map { |request_signature| Request.from_webmock_request_signature(request_signature) }
+        array
     end
 
     def to_s
--- a/lib/webmock/request_registry.rb
+++ b/lib/webmock/request_registry.rb
@@ -40,8 +40,7 @@ module WebMock
 
     def to_a
       requested_signatures.
-        hash.
-        flat_map { |request_signature, number_of_requests| [request_signature] * number_of_requests }.
+        array.
         map { |request_signature| Request.from_webmock_request_signature(request_signature) }
     end
 
--- a/lib/webmock/request_registry.rb
+++ b/lib/webmock/request_registry.rb
@@ -8,15 +8,16 @@ module WebMock
     attr_accessor :requested_signatures
 
     class Request
-      attr_accessor :method, :uri
+      attr_accessor :method, :uri, :headers
 
       def self.from_webmock_request_signature(request_signature)
-        new(method: request_signature.method, uri: request_signature.uri)
+        new(method: request_signature.method, uri: request_signature.uri, headers: request_signature.headers)
       end
 
-      def initialize(method:, uri:)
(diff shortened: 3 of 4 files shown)


lib/webmock/util/hash_counter.rb
```diff
@@ -5,22 +5,36 @@
...
+      def store_to_array(key:, num:)
+        request_object_id = @request_object_ids[key]
+        request_object_id = key.object_id if request_object_id.nil?
+        num.times do
+          array << ObjectSpace._id2ref(request_object_id)
+        rescue RangeError
+          # Points to an invalid or recycled object so ignore
+        end
+        @request_object_ids[key] = key.object_id
+      end
```
johngallagher, line 36: @bblimke this is missing any comments to explain what's going on - maybe worth me adding some?

For others - doing this so that we are just referencing existing requests in the array, rather than creating new objects.

This is so that if we have massive numbers of requests that have the same signature, we're not bloating memory.

johngallagher:
Gah @bblimke all the tests are failing *and* I realised I need to do some documentation (I'll do that in a separate PR if that's OK?)

I'll come back to this and debug.

johngallagher:
I'm getting a ton of unrelated CI failures locally - no idea what's going on here.

These tests are all failing on `master` branch:

image

Happy to pair with you some time @bblimke to make this all green again.

I'm going to leave this for now as I'm not best placed to do a deep dive into the codebase...

lib/webmock/request_signature.rb
```diff
@@ -44,6 +44,10 @@ def json_headers?
       !!(headers&.fetch('Content-Type', nil)&.start_with?('application/json'))
     end
 
+    def parsed_body
```
bblimke, line 47: @johngallagher  Thank you for this feature. I understand the intention behind this method. I have some concerns though. The current implementation assumes JSON, but request bodies can come in various formats (e.g., form-url-encoded, XML). This method would raise errors for non-JSON payloads. Additionally, while symbolized keys can be useful, it's often a matter of user preference rather than a universal expectation.

To address these issues, we could either expand `parsed_body` to support multiple formats, or allow users to apply their own decorators for custom parsing logic. Given these considerations, I suggest treating this as a separate feature. This would allow us to design a more flexible solution that caters to various use cases and preferences.

What are your thoughts on this approach?

lib/webmock/util/hash_counter.rb
```diff
@@ -5,22 +5,36 @@
 module WebMock
   module Util
     class HashCounter
-      attr_accessor :hash
+      attr_accessor :hash, :array
```
bblimke, line 8: @johngallagher I believe we should separate the array of requests from the HashCounter. The HashCounter has a specific, single responsibility, and adding an array of requests expands its scope beyond its intended purpose.

Instead, I suggest moving the ordered list of requests up to the RequestRegistry object. This way, the RequestRegistry would manage both the HashCounter and the array of requests, maintaining a clearer separation of concerns.

What are your thoughts on this restructuring? It would allow the HashCounter to remain focused on its core functionality.

lib/webmock/util/hash_counter.rb
```diff
@@ -5,22 +5,36 @@
...
           @order[key] = @max += 1
         end
       end
 
+      def store_to_array(key, num)
+        request_object_id = @request_object_ids[key]
+        request_object_id = key.object_id if request_object_id.nil?
+        num.times do
+          array << ObjectSpace._id2ref(request_object_id)
+        rescue RangeError
```
bblimke, line 32: @johngallagher have you experienced a scenario where that `rescue` is needed? when can that happen?

lib/webmock/util/hash_counter.rb
bblimke: @johngallagher Thanks for your work on this. I've been looking at the `store_to_array` method.

I think it might be helpful to add a comment explaining the intention behind using `object_id` and `_id2ref`. This would really help to understand the reasoning behind this approach.

That said, I believe we might be able to simplify this code. I'm not sure whether we really need `object_id` and `_id2ref` at all. From my understanding, `array` is just storing references to the original objects, not clones.

I think the following achieves the same result:

```ruby
def store_to_array(key, num)
  @request_objects ||= {}
  stored_object = @request_objects[key] ||= key
  num.times do
    array << stored_object
  end
end
```

This version uses a hash (`@request_objects`) to store the original request signature objects with same `#hash` result. It maintains the original object without the need for `object_id`.

It should be more efficient as it avoids the overhead of `ObjectSpace._id2ref` and potential `RangeError` rescues,
though I don't know how much faster that is.

Can you think of any scenarios where the original implementation behaves differently from this proposed one?

What are your thoughts on this?

I wonder if this can me optimised even further.

bblimke:
@johngallagher all tests in master branch are passing now, therefore feel free to marge master branch to this one.

lib/webmock/request_signature.rb
johngallagher: Yes, let's treat this as a separate feature. A massive overreach on my part! I'll rip it out.

lib/webmock/util/hash_counter.rb
johngallagher: @bblimke yup, that would work better. I do think that having all this cruft in the `RequestRegistry` might clutter it up, so it might be a `RequestStore` or something similar... although that's literally what `RequestRegistry` does I guess... so I'll look into that. Thanks! Great idea.

lib/webmock/util/hash_counter.rb
johngallagher: Let's take your advice and simplify this code! It can happen when an old reference is kept but it's out of bounds. This happened once or twice when running the test suite - I assumed it's because the references hadn't been cleaned up properly but never dug into it.

I didn't really like doing this, so anything that's simpler and doesn't use this weird `#_id2ref` method has my vote!

lib/webmock/util/hash_counter.rb
johngallagher: This is a **much** nicer solution than mine! I can't think of any scenarios just now but I'll think on it over the next few days...

lib/webmock/util/hash_counter.rb
johngallagher: @bblimke aha - I remember why I didn't put it into a separate object now. 💡 

It's because of the design - `HashCounter` is exposed directly in lots of places like so:

https://github.com/johngallagher/webmock/blob/088aa2678f7efb34e910e8b430b54980283b9485/lib/webmock/http_lib_adapters/curb_adapter.rb/#L61

Personally, I'd rather encapsulate how the registry stores requests and hide that from clients.

That would mean we could have one store, two or more. But given how it's done, without a huge refactoring, I don't see how we'd get this in.

I'd be happy to take on that refactoring, to be clear - I love me some refactoring!

It'd be:

```diff
- WebMock::RequestRegistry.instance.requested_signatures.put(request_signature)
+ WebMock::RequestRegistry.instance.put(request_signature)
```

And the `requested_signatures` would be entirely private. We could have two different store classes that way and do something like (sketching it out roughly):

```ruby
class WebMock::RequestRegistry
  def initialize
    @hash_counter_store = HashCounterStore.new,
    @array_store = ArrayStore.new
  end

  def put(request_signature)
    @hash_counter_store.put(request_signature)
    @array_store.put(request_signature)
  end

  def get(request_signature)
    @array_store.get(request_signature)
  end

  def to_a
    @array_store.to_a
  end

  def to_h
    @hash_counter_store.to_h
  end
  # ...
end
```

This would give us duck typing for stores, make their internals private and allow us to reach into the relevant store when we wanted to pluck out requests that were stored in a specific format.

WDYT?

lib/webmock/util/hash_counter.rb
johngallagher: OK, simplified as you suggested and it passes all the tests and seems to work! Nice.

lib/webmock/request_signature.rb
johngallagher: Ripped out.

lib/webmock/util/hash_counter.rb
johngallagher: @bblimke **nudge** :)

bblimke: @johngallagher Thanks for the reminder. I agree that RequestRegistry internals should be private, with no other object having direct access to the hash counter or the array. I like your solution with two separate stores: hash_counter_store and array_store.

I'm not sure if this is a breaking change, since RequestRegistry isn't supposed to be used directly. It's not part of WebMock::API, which is the only versioned interface. However, there's a likelihood that some people have accessed RequestRegistry directly.

I don't mind releasing a new major version though, if we decide it's necessary.

bblimke:
> I'm getting a ton of unrelated CI failures locally - no idea what's going on here.
> 
> These tests are all failing on `master` branch:

Tests on master branch is passing now, if you would like to rebase your branch.

PR #1070, opened 2024-09-22
Exclude em-http-request on Ruby 3.4

c960657:
The current version of em-http-request (1.1.7) is not compatible with Ruby HEAD (see https://github.com/igrigorik/em-http-request/pull/365).

Skip tests for the em-http-request integration on Ruby 3.4 for now in order to allow the other tests to complete.

Best reviewed with “ignore white-space”.

### DIFF
--- a/lib/webmock/http_lib_adapters/em_http_request_adapter.rb
+++ b/lib/webmock/http_lib_adapters/em_http_request_adapter.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-return if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('3.4.0') && Gem.loaded_specs['em-http-request'].version <= Gem::Version.new('1.1.7')
+return if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('3.4.0')
 
 begin
   require 'em-http-request'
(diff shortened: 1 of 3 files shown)


bblimke:
@c960657 Thanks for this PR! I believe the `skip` solution fits well for the temporary em-http-request issue.

However, I'm not sure the JRuby case should use `skip` as well. em-http-request has never supported and likely never will support Java. Therefore, I believe using `skip` will unnecessarily clutter RSpec output. The specs for em-http-request should never run on the Java platform in my opinion.

c960657:
Ok, I reversed that part.

spec/spec_helper.rb
```diff
@@ -3,7 +3,6 @@
 unless RUBY_PLATFORM =~ /java/
   require 'curb'
   require 'patron'
-  require 'em-http'
```
bblimke, line 6: @c960657 why is it important to remove that `require`?

c960657: On Ruby HEAD, even `require`'ing the library will raise because of the missing constant. Removing it here doesn't seem to affect the tests on the other Ruby versions AFAICT.

bblimke: @c960657 thank you for investigating that and explaining 👍

lib/webmock/http_lib_adapters/em_http_request_adapter.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+return if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('3.4.0')
```
matheussilvasantos, line 3: Could we have something like:

```
unless ENV["IGNORE_RUBY_VERSION_CHECK_FOR_EM"] == "enabled"
  return if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('3.4.0')
end
```
We've forked em-http-request where I work, so this isn't a problem.

PR #1078, opened 2024-11-06
Fix FrozenError in Typhoeus streaming response body

patrickjaberg:
When stubbing a response for the Typhoeus adapter, and the Typhoeus request has an `on_body` callback, a `FrozenError` exception is raised when attempting to concatenate the current chunk of the response to the existing response body (i.e. `response.body << chunk`).

FWIW, my use case for this is to abort a request as early as possible when the response body exceeds a given size, specifically when the response doesn't have a `Content-Length` header.

The example below illustrates the issue:

```ruby
require "bundler/inline"

gemfile do
  source "https://rubygems.org"

  gem "typhoeus", "1.4.1"
  gem "webmock", "3.24.0"
end

WebMock.enable!
WebMock.stub_request(:get, "https://example.com").to_return(status: "200", body: "body")

request = Typhoeus::Request.new("https://example.com")

request.on_body do |chunk, response|
  response.body << chunk
end

request.run
```

This change initializes the Typhoeus response body to a non-frozen, mutable string when using the `on_body` callback.

### DIFF
--- a/lib/webmock/http_lib_adapters/typhoeus_hydra_adapter.rb
+++ b/lib/webmock/http_lib_adapters/typhoeus_hydra_adapter.rb
@@ -172,7 +172,7 @@ if defined?(Typhoeus)
                 request.execute_headers_callbacks(response)
               end
               if request.respond_to?(:streaming?) && request.streaming?
-                response.options[:response_body] = ""
+                response.options[:response_body] = "".dup
                 request.on_body.each { |callback| callback.call(webmock_response.body, response) }
               end
               request.finish(response)
--- a/spec/acceptance/typhoeus/typhoeus_hydra_spec.rb
+++ b/spec/acceptance/typhoeus/typhoeus_hydra_spec.rb
@@ -125,6 +125,19 @@ unless RUBY_PLATFORM =~ /java/
           expect(test_complete).to eq("")
         end
 
+        it "should initialize the streaming response body with a mutible (non-frozen) string" do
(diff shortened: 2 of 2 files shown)


lib/webmock/http_lib_adapters/typhoeus_hydra_adapter.rb
```diff
@@ -172,7 +172,7 @@ def self.request_hash(request_signature)
                 request.execute_headers_callbacks(response)
               end
               if request.respond_to?(:streaming?) && request.streaming?
-                response.options[:response_body] = ""
+                response.options[:response_body] = String.new
```
koic, line 175: It's a corner case, using `"".dup` can reduce code size while maintaining behavioral compatibility.

```console
$ ruby -e 'p "".encoding'
#<Encoding:UTF-8>
$ ruby -e 'p "".dup.encoding'
#<Encoding:UTF-8>
$ ruby -e 'p String.new.encoding'
#<Encoding:BINARY (ASCII-8BIT)>
```

```suggestion
                response.options[:response_body] = "".dup
```

patrickjaberg: Thanks for your review & feedback @koic !

koic:
This looks good to me. Can you squash your commits into one and rebase onto the latest master branch?

patrickjaberg:
Apologies for the delay, @koic.  The changes have been rebased & squashed.  Thanks for your review and recent efforts on the webmock project to keep it moving forward!  🎉

koic:
Thanks!

PR #1083, opened 2024-11-21
stub_request accept Addressable::URI

notEthan:
stub_request accepts URI::Generic but not Addressable::URI


notEthan:
I'm usually good about checking for existing issues / PRs but I seem to have failed at that here. Closing for the earlier and better https://github.com/bblimke/webmock/pull/1073

PR #1084, opened 2024-11-22
Provide a default host header to make request as valid as possible

benoittgt:
While working on this https://github.com/sinatra/sinatra/pull/2053 in our project. I noticed than when using Webmock, Sinatra logs and especially the enforced [rack-protection](https://github.com/sinatra/sinatra/blob/main/rack-protection/lib/rack/protection/host_authorization.rb#L63-L69) were showing this kind of logs:

```
D, [2024-11-22T13:05:16.798156 #26673] DEBUG -- : Rack::Protection::HostAuthorization @all_permitted_hosts=[".company.com"] @permitted_hosts=["company.com"] @domain_hosts=[/\A(?-mix:[a-z0-9\-.]+)company\.com\z/i] @ip_hosts=[] origin_host="" forwarded_host=nil
```

As you can see, `origin_host` is empty, because the header is missing.

When not using webmock, we fallback on `net/http` host header setup. https://github.com/ruby/net-http/blob/cfbbb50c931a78fc2b5c731b9abeda161e1dfdd1/lib/net/http.rb#L2482

I think Webmock should also set this `host` header.

I looked at the test quickly but didn't find an easy to check for this change.


benoittgt:
Rebased. I think this pull request is still interesting.

Issue #1089, opened 2025-01-05
For `httprb` adapter, it should raise NetConnectNotAllowedError if connect is not allowed.


Novtopro:
If connect is not allowed, for httprb adapter, it will just return nil. I think we should raise a NetConnectNotAllowedError just like other adapters.

PR #1090, opened 2025-01-05
#1089 Raise error if connect is not allowed for httprb

Novtopro:


koic:
This change seems reasonable. Can you add a regression test?

PR #1091, opened 2025-01-10
Fix included modules on Ruby 3.4

sailor:
We currently get this error when trying to use webmock with Ruby 3.4:

```
Failure/Error: require File.expand_path('../config/environment', __dir__)

NameError:
  undefined method 'do_get_block' for class 'WebMockHTTPClient'
# /gems/webmock-3.24.0/lib/webmock/http_lib_adapters/httpclient_adapter.rb:233:in 'Module#alias_method'
```

It seems that it's defining aliases from method that are not included yet is not supported anymore by ruby >= 3.4.

This patch fixes the problem by just including the module before defining the aliases.


knapo:
@bblimke any chance we have this merged and released?  Dzięki!

koic:
This PR was reverted in #1095 due to failing CI:
https://github.com/bblimke/webmock/actions/runs/13170312401/job/36759323715

If possible, I would appreciate it if you could resolve the issue in a way that allows the CI to pass.

knapo:
Thanks @koic! This change isn't actually necessary since webmock is fully compatible with Ruby 3.4. The real issue is downstream: httpclient fails to load in Ruby 3.4 when the mutex_m gem isn't present (it's no longer included in Ruby 3.4.0's default gems). This creates a situation where defined?(::HTTPClient) returns true, but HTTPClient isn't properly loaded."

PR #1092, opened 2025-01-11
fix: make specs pass with ruby 3.4

Juanmcuello:
The hash serialization to string seems to be different in ruby 3.4:

$ chruby 3.3
$ ruby -e 'puts ({"a" => 1}).to_s'
{"a"=>1}

$ chruby 3.4
$ ruby -e 'puts ({"a" => 1}).to_s'
{"a" => 1}

This change makes specs to use the serialized string instead of the hardcoded string, so the output is now independent of the way the string is serialized.


Juanmcuello:
@bblimke , just to know if you had a chance to take a look at this. Thanks!

PR #1099, opened 2025-02-10
Add :any parameter for query options

### LINKED ISSUE #877: Stubbing a request with any query
Currently, when I stub a request with any query, below code.

```ruby
stub_request(:get, 'example.com/search')
  .with(query: hash_including({}))
  .and_return(status: 200)
```

I suggest below interface. How do you think? If you think it is not sure, please feel free to close.

```ruby
stub_request(:get, 'example.com/search')
  .with(query: :any)
  .and_return(status: 200)
```

## Reference

- https://github.com/bblimke/webmock/issues/693

nightpool:
Implements and closes #877!

Implementation was very straightforward, the only doubt I had was the first parameter to `AnyArgMatcher`. I'm not sure if the `nil` argument is necessary, since we explicitly ignore it. Feels like it would be better to make initialize allow a default value of nil.

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,6 +278,14 @@ Net::HTTP.get('www.example.com',
   '/thing/5.json?x=1&y=2&z=3&anyParam=4')    # ===> Success
 ```
 
+### Allow query params
+
+```ruby
+stub_request(:get, "www.example.com").with(query: :any)
+
+RestClient.get("http://www.example.com/?q=example")    # ===> Success
+```
+
 ### Matching query params using hash
 
 ```ruby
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -278,6 +278,14 @@ Net::HTTP.get('www.example.com',
   '/thing/5.json?x=1&y=2&z=3&anyParam=4')    # ===> Success
 ```
 
+### Allow query params
+
+```ruby
+stub_request(:get, "www.example.com").with(query: :any)
```
nightpool, line 284: Even this usage still feels a little heavy-weight to me.... Is there any reason we can't do `stub_request(:get, "www.example.com", query: :any)`? Or even just make it the default?

PR #1103, opened 2025-08-04
Return support for `em_http_request`

Koilanetroc:
In version 3.24.0 support for `em_http_request` was dropped due to:
> The current version of em-http-request (1.1.7) is incompatible with Ruby 3.4 due to an unresolved issue (https://github.com/igrigorik/em-http-request/pull/365). Support for em-http-request will be re-enabled once the compatibility issue is resolved.

The problem was addressed in https://github.com/igrigorik/em-http-request/issues/354 and now cookiejar supports ruby 3.3+, it was added in [this commit](https://github.com/dwaite/cookiejar/commit/36332870693334ad9fb51dbfebc398b46bb1427a).

Therefore I'm returning back support for `em_http_request` which was disabled in https://github.com/bblimke/webmock/pull/1070

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -2,6 +2,7 @@ require 'rubygems'
 require 'httpclient'
 unless RUBY_PLATFORM =~ /java/
   require 'curb'
+  require 'em-http'
   require 'patron'
   require 'typhoeus'
 end
@@ -49,4 +50,3 @@ RSpec.configure do |config|
 
   config.include Failures
 end
-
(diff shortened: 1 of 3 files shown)


Koilanetroc:
Seems like currently something is off with CI, I tested a dummy change(only changed a line in readme) and CI was red anyway, also failures seems to be unrelated to my changes.

matheussilvasantos:
It would be nice to get this merged.

spec/spec_helper.rb
```diff
@@ -4,6 +4,7 @@
   require 'curb'
   require 'patron'
   require 'typhoeus'
+  require 'em-http'
```
koic: Ah, that’s my two cents. Can you reorder it as follows?

```suggestion
  require 'curb'
  require 'em-http'
  require 'patron'
  require 'typhoeus'
```

koic:
Thanks!

matheussilvasantos:
@koic, do you know when releases to Ruby Gems usually happen?

koic:
Yeah, I plan to release it today or tomorrow.

koic:
WebMock 3.25.2 has been released.

PR #1109, opened 2025-10-27
Fix tests compatibility with recent async-http versions

misdoro:
Since https://github.com/socketry/protocol-http/pull/84 ([protocol-http 0.51.0](https://github.com/socketry/protocol-http/blob/main/releases.md#v0510)), internal dependency of async-http gem
the tests were failing complaining about a duplicate `content-length` header.

CI: allow running tests on all push events, useful for running tests before opening a PR

- Remove the version pin of protocol-http gem
- Split the request and response sample headers in the tests
- explicitly remove the stubbed content-length response header in async-http protocol adapter

Follow-up of #1107


bblimke:
@Earlopain thank you! Just pushed it.

misdoro:
@bblimke it was the last one :sweat_smile: , I won't bother you any more :angel:

PR #1111, opened 2025-11-01
confirmed it works with curb 1.2.2

taf2:
Did some testing and confirm this gem works with the latest version's of curb.


taf2:
looks like the CI just failed on 3.0... maybe re-run and it passes?

PR #1121, opened 2026-03-12
[fix] add support to parse http/2 request on curb adapter

criess:
+ some test for CurbAdapters parse_header_string class level API

I came across it when letting it play with vcr, then somehow vcr recorded strange headers so I made a case to support HTTP/2 on a header string being parsed.

### DIFF
--- a/spec/acceptance/curb/curb_spec.rb
+++ b/spec/acceptance/curb/curb_spec.rb
@@ -507,4 +507,32 @@ unless RUBY_PLATFORM =~ /java/
       end
     end
   end
+
+  describe WebMock::HttpLibAdapters::CurbAdapter do
+    let(:header_tpl) { "HTTP/_VERSION_ 500 BROKEN\r\nX-anyheader: anyvalue" }
+
+    describe '.parse_header_string' do
+      %w[1.0 1.1 2].each do |http_v|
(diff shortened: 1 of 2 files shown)


spec/acceptance/curb/curb_spec.rb
```diff
@@ -507,4 +507,27 @@
...
+        it "returns correct status from HTTP/#{http_v}" do
+          input = header_tpl.sub("_VERSION_", http_v)
+          expect(described_class.parse_header_string(input)[0]).to eq("BROKEN")
+        end
+      end
+
+      it "returns correct single line of header (HTTP version known)" do
+        input = header_tpl.sub("_VERSION_", "1.1")
+        expect(described_class.parse_header_string(input)[1]).to match(Hash["X-anyheader", "anyvalue"])
+      end
```
koic: Since HTTP/2 omits the reason phrase, a test case for a status line without a reason phrase (e.g., `HTTP/2 500`) should also be added.

```suggestion
      end

      it "returns correct status from HTTP/2 without reason phrase" do
        input = "HTTP/2 500\r\nX-anyheader: anyvalue"                                                                                                                              
        expect(described_class.parse_header_string(input)[0]).to eq("")                                                                                                            
      end 
```

PR #1123, opened 2026-03-17
Add HTTP.rb v6.0.0 compatibility to http_rb adapter

sferik:
HTTP.rb 6.0.0 was just released with [several breaking changes](https://github.com/httprb/http/blob/main/CHANGELOG.md#600---2026-03-16) that affect the WebMock adapter:
* HTTP verb methods require keyword arguments
* `Response.new` takes keyword arguments
* `Stream#readpartial` raises `EOFError` at EOF instead of returning `nil`
* Features are wrapped in reverse order during response building

All these fixes are compatible back to v0.8, the minimum supported version in the gemspec.


koic:
WebMock 3.26.2 has been released. Thank you!
https://github.com/bblimke/webmock/releases/tag/v3.26.2

PR #1125, opened 2026-04-02
Expose Async::HTTP::Client#pool for async-websocket (and puppeteer-ruby) compatibility

s-mage:
Really this PR tries to allow puppeteer-ruby to connect to the browser via websocket. It happens to use `Async::HTTP::Client` that webmock patches for that, but uses pool directly, acquires connection from it and runs the request on it.

Code that I'm trying to run is essentially
```
Puppeteer.launch(headless: true, args: ["--no-sandbox"]) do |browser|
  page = browser.pages.first || browser.new_page
  page.goto("http://google.com", wait_until: "networkidle0")
end
```

with `puppeteer-ruby (0.51.0)`

This is a PoC to see if people who know way more than me say if it's a webmock or async-websocket problem at all. If it's the right place, I'll add test and cleanup comments somewhat. 

Thanks!

### DIFF
--- a/lib/webmock/http_lib_adapters/async_http_client_adapter.rb
+++ b/lib/webmock/http_lib_adapters/async_http_client_adapter.rb
@@ -92,6 +92,21 @@ if defined?(Async::HTTP)
           @webmock_client.close
         end
 
+        # WebMock doesn't mock connection pools. If your code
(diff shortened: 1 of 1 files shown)


s-mage:
I realised it's a wrong thing to do and I should instead disable webmock before running puppeteer and enable it back when done. Sorry for the fuss!

PR #1126, opened 2026-04-09
Add proxy matching to adapters that support it.

nialbima:
This is a very-brainless implementation of matching on `proxy` for Webmock. Not all the supported Webmock adapters supported proxy config in an easy way, so I implemented these ones only (the other ones ignore the argument):
- Net::HTTP
- Excon
- Patron
- Typhoeus
- Curb

It's really, really similar to how Webmock implements headers. I added examples to the README, and made sure to cover my work thoroughly (which is why this PR is 500 lines).

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Features
 
 * Stubbing HTTP requests at low http client lib level (no need to change tests when you change HTTP library)
 * Setting and verifying expectations on HTTP requests
-* Matching requests based on method, URI, headers and body
+* Matching requests based on method, URI, headers, body and proxy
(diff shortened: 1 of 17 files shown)


Nakilon:
> a very-brainless

because Claude? /s

nialbima:
@Nakilon oh absolutely. I reviewed all of the work myself and it's really, really similar to how headers were implemented here, so I felt comfortable opening the PR. But yeah, this was "i need my app to confirm that it actually found the proxy i need it to find, Clerd, have at it."

PR #1127, opened 2026-05-24
Confirm to work with curb 1.3.7

jbockler:
The 1.3.x line is primarily Ruby 4.0 support plus Curl::Multi/Fiber-scheduler bugfixes, no public API of Curl::Easy changed, so the WebMock adapter is unaffected.

### DIFF
--- a/lib/webmock/http_lib_adapters/curb_adapter.rb
+++ b/lib/webmock/http_lib_adapters/curb_adapter.rb
@@ -7,7 +7,7 @@ rescue LoadError
 end
 
 if defined?(Curl)
-  WebMock::VersionChecker.new('Curb', Curl::CURB_VERSION, '0.7.16', '1.2.2', ['0.8.7']).check_version!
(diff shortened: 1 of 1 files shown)


jbockler:
The 2.6 error seems to be pre-existing. Maybe consider just dropping support for it, as it is EOL since 2022.

koic:
Libraries are used in a much wider variety of ways than applications. Because of that, the view here is that libraries should support older runtimes as much as reasonably possible, independently of the runtime's official EOL status.

For example, one possible approach would be to split the existing and new code paths with something like `RUBY_VERSION >= '2.7'`. This does not significantly harm maintainability.

jbockler:
@koic the curb bug on ruby 2.6 was fixed in curb 1.3.6 (see taf2/curb#478)
I updated the version checker to allow 1.3.x, but kept a warning for the affected versions on Ruby 2.6.

koic:
@jbockler Could you update the title to "Confirm to work with curb 1.3.7" and squash the commits into one?

jbockler:
> @jbockler Could you update the title to "Confirm to work with curb 1.3.7" and squash the commits into one?

Updated 👍

koic:
Thanks!

Issue #711, opened 2017-07-06
random failure of spec/acceptance/httpclient/httpclient_spec.rb[1:5:1:4:4]


lnussbaum:
Hi,

spec/acceptance/httpclient/httpclient_spec.rb[1:5:1:4:4] fails randomly with the following error:
```
  1) HTTPClient when using async requests with WebMock when request stub declares that request should raise exception should raise exception after returning declared successful response first
     Failure/Error: raise @exception if @exception

     MyException:
       Exception from WebMock
     Shared Example Group: "declared responses" called from ./spec/acceptance/webmock_shared.rb:29
     Shared Example Group: "with WebMock" called from ./spec/acceptance/httpclient/httpclient_spec.rb:46
     # ./lib/webmock/response.rb:66:in `raise_error_if_any'
     # ./lib/webmock/http_lib_adapters/httpclient_adapter.rb:118:in `build_httpclient_response'
     # ./lib/webmock/http_lib_adapters/httpclient_adapter.rb:61:in `do_get'
     # ./lib/webmock/http_lib_adapters/httpclient_adapter.rb:51:in `do_get_stream'

Finished in 7.31 seconds (files took 0.83351 seconds to load)
627 examples, 1 failure

Failed examples:

rspec ./spec/acceptance/httpclient/httpclient_spec.rb[1:5:1:4:4] # HTTPClient when using async requests with WebMock when request stub declares that request should raise exception should raise exception after returning declared successful response first
```

To reproduce:
```
while rspec ./spec/acceptance/httpclient/httpclient_spec.rb ; do :; done
```

lnussbaum:
or:
```
  1) HTTPClient when using async requests with WebMock when requests are stubbed when stubbing request with a global hook should call the block only once per request
     Failure/Error: expect(call_count).to eq(1)

       expected: 1
            got: 2

       (compared using ==)
     Shared Example Group: "stubbing requests" called from ./spec/acceptance/webmock_shared.rb:27
     Shared Example Group: "with WebMock" called from ./spec/acceptance/httpclient/httpclient_spec.rb:46
     # ./spec/acceptance/shared/stubbing_requests.rb:499:in `block (4 levels) in <top (required)>'

Finished in 7.6 seconds (files took 0.81887 seconds to load)
627 examples, 1 failure

Failed examples:

[...]

PR #1131, opened 2026-08-21
Fix race condition in `HTTPClient` adapter async requests

koic:
An async request could evaluate its stub twice, consuming an additional response from the stub sequence, so stubs declared with `to_raise(...).times(2).to_return(...)` could skip an expected exception or return sequenced responses out of order. This surfaced as a flaky CI failure:
https://github.com/bblimke/webmock/actions/runs/32505353347/job/96844160056
    
Two races led to the same extra evaluation. The thread spawned by `HTTPClient#do_request_async` could start running before the caller copied WebMock's thread variables into it. And since building a request signature calls `headers_from_session`, which stamps a `Date` header with 1 second resolution, the signature the async thread rebuilt in `do_get` could stop matching the one the response was cached under once the clock crossed a second boundary, as caught on CI after the first attempt fixed only the thread startup race:
https://github.com/koic/webmock/actions/runs/32519734490/job/96889126317
    
Build the signature and evaluate the stub once, on the calling thread, and hand both to a thread the adapter spawns itself, which only delivers the prepared response instead of going through `do_get`. Real requests with net connect allowed keep using HTTPClient's original implementation, where the late thread variable copy is harmless because evaluating an unmatched request does not consume anything. The regression test delays the first request beyond one second so both races are reproduced deterministically.
    
This is not limited to WebMock's own CI. The races have shipped in releases since 3.22.0, so users who stub `HTTPClient` async requests with response sequences could hit the same intermittent failures in their test suites.
    
The randomly failing example reported in #711 shows the same symptom, an exception declared for a later response being raised on the first async request, and with the current implementation that is exactly what these races cause.


bblimke (approved):
Thank you @koic !

koic:
After approval, CI on my fork caught the new regression test failing on Ruby 3.1: https://github.com/koic/webmock/actions/runs/32519734490/job/96889126317

There was a second race with the same symptom: the async thread rebuilt the request signature in `do_get`, and since `headers_from_session` stamps a `Date` header with 1-second resolution, a rebuild across a second boundary missed the cached response and consumed an extra stubbed one.

I pushed a commit that evaluates the stub once on the calling thread and hands the prepared response to the async thread, removing the signature rebuild, the thread-variable handoff, and the queue handshake. The regression test now delays the first request beyond one second to cover both races deterministically.

bblimke (approved):
Thank you again!

